PR #3, opened 2020-05-19
Word

aleksander-mendoza:
Added draft of my idea how word could be implemented

### DIFF
--- a/src/Word.jl
+++ b/src/Word.jl
@@ -1,0 +1,68 @@
+abstract type Word{T} <: AbstractVector{T} end
+
+
+
+#interface for morphism T->Int
+function to_int{T}(t::T) end
+#interface for morphism Int->T
+function from_int{T}(t::T) end
+#interface for ordering on T
+function <=(a::T,b::T) end
+#empty string as monoid neutral element
+Base.one(::Type{Word{T}}) where {T}  = Word(Vector{T}())
+#overload for <= that implements word ordering
+function <=(a::Word{T},b::Word{T})
+#overload for * as concatenation
+function *(a::Word{T},b::Word{T}) where {T} Word(a.ptrs+b.ptrs) end
+#interface for function that returns actual vector inside word
+function data(a::Word{T})
+
+Base.length(w::Word{T}) where {T}  = length(w.ptrs)
+
+
+#Is there function composition in Julia?
+#Is it possible to assign function more or less like this?
+#In Haskell I could just make it work
+function Base.iterate(w::Word{T}) = data(w) . iterate()
+
+struct WordLexLen{T}<:Word{T}
+    ptrs::Vector{T}
+end
+
+struct WordLex{T}<:Word{T}
+    ptrs::Vector{T}
+end
+
+#Julia does not have dependet types, but type constructors
+#can be emulated by manually specyfing all inhabitants of type family.
+#In this case the type family is {WordLexLen,WordLex}
+
+#This can be though of as one of the "axioms" for Word
+function data(a::WordLexLen{T})
+  return a.ptrs
+end
+
+function data(a::WordLex{T})
+  return a.ptrs
+end
+
+#This can be though of as another "axiom" for Word
+function <=(a::WordLexLen{T},b::WordLexLen{T})
+  if length(a) < length(b) return true end
+  if length(a) > length(b) return false end
+  for (char_in_a,char_in_b) in zip(a,b)
+    if char_in_a > char_in_b
+      return false
+    end
+  end
+  return true
+end
+
+function <=(a::WordLex{T},b::WordLex{T})
+  for (char_in_a,char_in_b) in zip(a,b)
+    if char_in_a > char_in_b
+      return false
+    end
+  end
+  return true
+end


src/Word.jl
```diff
@@ -0,0 +1,68 @@
...
+#interface for function that returns actual vector inside word
+function data(a::Word{T})
+
+Base.length(w::Word{T}) where {T}  = length(w.ptrs)
+
+
+#Is there function composition in Julia?
+#Is it possible to assign function more or less like this?
+#In Haskell I could just make it work
+function Base.iterate(w::Word{T}) = data(w) . iterate()
```
kalmarek, line 26: there is composition, but here we need to define special methods:
```julia
Base.iterate(w::Word{T}) = iterate(w.ptrs)
Base.iterate(w::Word{T}, state) = iterate(w.ptrs, state)
```
is enough

src/Word.jl
```diff
@@ -0,0 +1,68 @@
...
+#empty string as monoid neutral element
+Base.one(::Type{Word{T}}) where {T}  = Word(Vector{T}())
+#overload for <= that implements word ordering
+function <=(a::Word{T},b::Word{T})
+#overload for * as concatenation
+function *(a::Word{T},b::Word{T}) where {T} Word(a.ptrs+b.ptrs) end
+#interface for function that returns actual vector inside word
+function data(a::Word{T})
+
+Base.length(w::Word{T}) where {T}  = length(w.ptrs)
```
kalmarek, line 20: `length` is fine, but for vectors/matrices/nd-arrays it's better to implement `Base.size(w::Word) = size(w.ptrs)`

kalmarek: in general have a look at https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array-1
it may seem a little bit cryptic, at first but all it says is that for a subtype of `AbstractArray` we need to implement
`size`, `getindex`, `setindex` (both with the appropriate number of args), and `IndexStyle` → `IndexLinear()` for it to behave like a vector.

most of the standard stuff (like e.g. iteration) will work automatically

src/Word.jl
kalmarek: note that it is not necessary to implement those if array interface is implemented correctly

src/Word.jl
```diff
@@ -0,0 +1,68 @@
...
+struct WordLex{T}<:Word{T}
+    ptrs::Vector{T}
+end
+
+#Julia does not have dependet types, but type constructors
+#can be emulated by manually specyfing all inhabitants of type family.
+#In this case the type family is {WordLexLen,WordLex}
+
+#This can be though of as one of the "axioms" for Word
+function data(a::WordLexLen{T})
```
kalmarek, line 41: one could use `a::Union{WordLexLen, Wordlex}` here.
Note that as long you don't need the actual type `T` there is no need to specify it.
(ok, sometimes it is needed, but not here)

kalmarek: also, with correctly defined array interface hopefully we should never use `data`, i.e. direct access to the storage of a word :)

src/Word.jl
```diff
@@ -0,0 +1,68 @@
...
+  return a.ptrs
+end
+
+function data(a::WordLex{T})
+  return a.ptrs
+end
+
+#This can be though of as another "axiom" for Word
```
kalmarek, line 49: these should be handled by the ordering

src/Word.jl
```diff
@@ -0,0 +1,68 @@
+abstract type Word{T} <: AbstractVector{T} end
+
+
+
+#interface for morphism T->Int
+function to_int{T}(t::T) end
```
kalmarek, line 6: in julia there is usually no need to specify those functions. Well written libraries should handle any type of `Integers`

src/Word.jl
```diff
@@ -0,0 +1,68 @@
...
+#interface for morphism Int->T
+function from_int{T}(t::T) end
+#interface for ordering on T
+function <=(a::T,b::T) end
+#empty string as monoid neutral element
+Base.one(::Type{Word{T}}) where {T}  = Word(Vector{T}())
+#overload for <= that implements word ordering
+function <=(a::Word{T},b::Word{T})
+#overload for * as concatenation
+function *(a::Word{T},b::Word{T}) where {T} Word(a.ptrs+b.ptrs) end
```
kalmarek, line 16: here we need something more:
```julia
Base.:*(a::Word{T}, b::Word{T}) where T = Word([a.ptrs; b.ptrs]) # syntax for appending
```
the syntax for overloading `*` (it needs to be prepended with `:`) is a little cumbersome

src/Word.jl
```diff
@@ -0,0 +1,68 @@
+abstract type Word{T} <: AbstractVector{T} end
```
kalmarek, line 1: In general one would write a (markdown formatted) docstring above the definition
```julia
"""
    Word{T}
Short description

Longer description
The subtypes of `Word{T}` need to implement the following methods
 * `Base.one` the monoid identity element (empty word)
 * `Base.:*` for monoid binary operation (concatenation)
 * iteration protocol
 * `length` the length of word in the appropriate word-length metric
....
"""
abstract type Word{T} <:AbstractVector{T}
```

this way when you type `?Word` in REPL You will see the above docstring

src/Word.jl
kalmarek: and this is an old syntax (pre julia-1.0), now it should be written as
```julia
function to_int(t::T) where T
    ...
end
```
or as a one-liner:
```julia
to_int(t::T) where T = ...
```

kalmarek:
how is it going @aleksander-mendoza ?

kalmarek:
superseded by #3

PR #4, opened 2020-05-19
Alphabet --- First attempt

bprzybylski:
This is the first attempt on the alphabet implementation.

### DIFF
--- a/src/Alphabet.jl
+++ b/src/Alphabet.jl
@@ -1,0 +1,161 @@
+import Base: push!
+
+"""
+    mutable struct Alphabet{T}
+
+A basic struct for storing alphabets. An alphabet consists of the symbols of a common type `T`.
+
+# Example
+```julia-repl
+julia> Alphabet{String}()
+Empty alphabet of String
+```
+"""
+mutable struct Alphabet{T}
+    alphabet::Vector{T}
+    inversions::Vector{Int}
+    function Alphabet{T}() where T
+        new(Vector{T}[], Vector{Int}[])
+    end
+end
+
+function Base.show(io::IO, A::Alphabet{T}) where T
+    if length(A.alphabet) == 0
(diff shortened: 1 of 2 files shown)


src/Alphabet.jl
```diff
@@ -0,0 +1,161 @@
...
+Alphabet of String:
+    1. "a" = ("c")⁻¹
+    2. "b"
+    3. "c" = ("a")⁻¹
+
+julia> get_symbol(A, -get_index(A, "a"))
+"c"
+```
+"""
+function get_symbol(A::Alphabet{T}, p::Int) where T
```
kalmarek, line 154:
```suggestion
function Base.get_index(A::Alphabet{T}, p::Integer) where T
```
slightly more general signature and easier access, through standard brackets

src/Alphabet.jl
```diff
@@ -0,0 +1,161 @@
...
+    if p > 0
+        return A.alphabet[p]
+    elseif p < 0 && A.inversions[-p] > 0
+        return A.alphabet[A.inversions[-p]]
+    end
+    nothing
```
kalmarek, line 160:
```suggestion
   throw(DomainError("Inverse to $(A.alphabet[-p]) is not defined"))
```

src/Alphabet.jl
bprzybylski: Yes, but the function does not return an index, so why to change the name to `get_index`?

kalmarek: because then we can write `A[1]` for the first symbol and `A[-2]` for the inverse of the second

kalmarek: oh, and of course it's `Base.getindex` :P

bprzybylski: That is so clever!

PR #5, opened 2020-05-20
Argument list correction + Alphabets of integers

bprzybylski:
This is a small correction of the previous pull request changes. It does not influence anything, but I wanted the types to be consistent.

**Edit.** I have extended this pull request by support for alphabets of integers. Previously, we had problems with the `getindex` function (one could not find a symbol in the alphabet of integers). Now, there is a lower-level function `getindexbysymbol` that explicitly finds a symbol, whatever type it is (including `Integer`).

### DIFF
--- a/src/Alphabet.jl
+++ b/src/Alphabet.jl
@@ -111,7 +111,7 @@ function set_inversion!(A::Alphabet{T}, x::T, y::T) where T
 end
 
 """
-    function getindexbyelement(A::Alphabet{T}, x::T) where T
+    function getindexbysymbol(A::Alphabet{T}, x::T) where T
 
 Returns the position of the symbol `x` in the alphabet `A`.
 
--- a/src/Alphabet.jl
+++ b/src/Alphabet.jl
@@ -111,9 +111,36 @@ function set_inversion!(A::Alphabet{T}, x::T, y::T) where T
 end
 
 """
-    function Base.getindex(A::Alphabet{T}, x::T) where T
+    function getindexbyelement(A::Alphabet{T}, x::T) where T
 
 Returns the position of the symbol `x` in the alphabet `A`.
+
+# Example
+```julia-repl
+julia> A = Alphabet{String}()
+Empty alphabet of String
+
+julia> push!(A, "a", "b", "c")
+Alphabet of String:
+    1. "a"
+    2. "b"
+    3. "c"
+
+julia> getindexbyelement(A, "c")
+3
+```
+"""
+function getindexbyelement(A::Alphabet{T}, x::T) where T
+    if (index = findfirst(symbol -> symbol == x, A.alphabet)) == nothing
+        throw(DomainError("Element '$(x)' not found in the alphabet"))
+    end
+    index
+end
+
+"""
+    function Base.getindex(A::Alphabet{T}, x::T) where T
+
+Returns the position of the symbol `x` in the alphabet `A`. If `T` is equal to `Integer`, use `getindexbyelement`.
 
 # Example
 ```julia-repl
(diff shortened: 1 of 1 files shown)


kalmarek:
hmm... I'm not sure that this fixes anything; I think it exposes a flaw that I didn't think about: consider what will happen if we have `A::Alphabet{Int}`:
* `A[1]` gives you the index of "1" in the alphabet, but 
* `A[Int32(1)]` gives you the first letter of the alphabet
Both are `Int` and the user will never know which one is which.

I think it's best to forbid `Integers` as letters for now?

bprzybylski:
> hmm... I'm not sure that this fixes anything; I think it exposes a flaw that I didn't think about: consider what will happen if we have `A::Alphabet{Int}`:
> 
> * `A[1]` gives you the index of "1" in the alphabet, but
> * `A[Int32(1)]` gives you the first letter of the alphabet
>   Both are `Int` and the user will never know which one is which.
> 
> I think it's best to forbid `Integers` as letters for now?

Well, not really. Consider the following example.

```julia-repl
julia> A = Alphabet{Int}()
Empty alphabet of Int64

julia> push!(A, 7, 3 ,5)
Alphabet of Int64:
    1.  7
    2.  3
    3.  5

julia> A[1]
7

julia> A[Int32(1)]
7
```

The reason to introduce `getindexbysymbol` is that by default, if the index is an integer, it is interpreted as the position. Whatever subtype you use.

kalmarek:
I actually didn't understand why
```julia
getindex(A::KnuthBendix.Alphabet{T}, p::Integer) where T
```
is more specific than
```julia
getindex(A::KnuthBendix.Alphabet{T}, x::T) where T
```
(it is because `T<:Any` in the first argument and `Int32<:Integer` is more specific than `Int32<:Any` in the second, one can learn something each day:).

I fully agree with the reason of introducing `getindexbysymbol`, but still: the result is that if we want to support integers as "letters" we can't call `A[letters[2]]` as its meaning is determined by the type of a letters (it will be either index of a letter, or the letter itself):
```julia
julia> let letters = ["7","1","5"]
           A = KnuthBendix.Alphabet{eltype(letters)}()
           push!(A, letters...)
           @show A[letters[2]]
       end;
A[letters[2]] = 2 # the index of second letter

julia> let letters = [7, 1, 5]
           A = KnuthBendix.Alphabet{eltype(letters)}()
           push!(A, letters...)
           @show A[letters[2]]
       end;
A[letters[2]] = 7 # the first letter
```

bprzybylski:
Yes, so if you know that your symbols are integers, you need to use `getindexbysymbol` explicitly. Earlier, there was no chance to do that kind of search.

To be precise, this update did not change anything to how non-integer alphabets are managed. It just introduced a way to do the same things with integer alphabets. I thing this is a "dobra zmiana", at least for now, because now all the possible cases are covered.

kalmarek:
I agree with the point that `getindexbysymbol(A, l)` gives us more freedom, but the issue of double meaning still persist: We can't express (at the moment) the constraint that `T` is not an `Integer`;
Therefore everywhere in our library code we can't use `A[letters[i]]` because the meaning depends on what is the `eltype` of `letters`.

To keep our general methods easy to write I'd make this assumption that `T` is not an integer enforcing it in the constructor of `Alphabet`. Unless we want to write `getindexbysymbol` all the time (I personally don't want to).

bprzybylski:
Okay. Now you can't create the alphabet of integers unless you explicitly state that you know what you are doing. Is that an acceptable compromise?

kalmarek:
I want to use the `getindex` syntax throughout the library code, so then the user is on him/her own ;)
Fair enough.

PR #6, opened 2020-05-24
Words

kalmarek:


src/words.jl
```diff
@@ -0,0 +1,83 @@
...
+Return the inverse of given word by reversing and negating its pointers.
+"""
+function Base.inv(w::Word)
+    res = similar(w)
+    n = length(w)
+    for (i, l) in enumerate(w)
+        res[n+1-i] = -l
+    end
+    return res
+end
```
bprzybylski, line 83: Is there any particular reason why you didn't write something like the following?
```julia
function Base.inv(w::Word)
    res = copy(w)
    reverse!(res.ptrs)
    return -1 * res
end
```

kalmarek: two additional allocations ;)

bprzybylski: Okay, one by calling the `copy` function and the second one to return `-1 * res`. Still, do you think that an explicit `for` loop is the most optimal approach? If yes, then okay.

kalmarek:
Let's check:
```julia
julia> using BenchmarkTools

julia> function inv2(w::KnuthBendix.Word)
           res = copy(w)
           reverse!(res.ptrs)
           return -1 * res
       end;

julia> inv3(w::KnuthBendix.Word) = KnuthBendix.Word(-reverse(w.ptrs));

julia> w = KnuthBendix.Word(1:100);

julia> @benchmark inv($w)
BenchmarkTools.Trial: 
  memory estimate:  304 bytes
  allocs estimate:  2
  --------------
  minimum time:     75.035 ns (0.00% GC)
  median time:      83.895 ns (0.00% GC)
  mean time:        101.736 ns (9.08% GC)
  maximum time:     3.247 μs (94.33% GC)
  --------------
  samples:          10000
  evals/sample:     940

julia> @benchmark inv2($w)
BenchmarkTools.Trial: 
  memory estimate:  1.17 KiB
  allocs estimate:  3
  --------------
  minimum time:     301.936 ns (0.00% GC)
  median time:      326.307 ns (0.00% GC)
  mean time:        386.202 ns (7.04% GC)
  maximum time:     7.656 μs (93.20% GC)
  --------------
  samples:          10000
  evals/sample:     249

julia> @benchmark inv3($w)
BenchmarkTools.Trial: 
  memory estimate:  592 bytes
  allocs estimate:  3
  --------------
  minimum time:     211.913 ns (0.00% GC)
  median time:      231.745 ns (0.00% GC)
  mean time:        265.233 ns (5.79% GC)
  maximum time:     3.730 μs (60.68% GC)
  --------------
  samples:          10000
  evals/sample:     530

julia> @assert inv(w) == inv2(w) == inv3(w)

```
When I last checked enumerate doesn't play well with auto-vectorisation, so maybe rewriting `inv` as
```julia
function Base.inv(w::Word)
    res = similar(w)
    n = length(w)
    for i in eachindex(w)
        @inbounds res[n+1-i] = -w[i]
    end
    return res
end
```
will make it faster (but at this point this is pre-matrue optimization ;)

---
loops in julia are as efficient as they are in `C` (with `@inbounds`). When llvm manages to emit vectorized asm for you, they can become even much faster

bprzybylski:
Thai is so cool to know! Thanks!

kalmarek:
just a follow up: with the version of `inv` above looking at `@code_native inv(w)` you can see
```asm
; │┌ @ int.jl:51 within `-'
L304:
        vpsubw  -96(%rsi), %ymm0, %ymm2
        vpsubw  -64(%rsi), %ymm0, %ymm3
        vpsubw  -32(%rsi), %ymm0, %ymm4
        vpsubw  (%rsi), %ymm0, %ymm5
; │└
; │┌ @ words.jl:69 within `setindex!' @ array.jl:826
        vpshufb %ymm1, %ymm2, %ymm2
        vpermq  $78, %ymm2, %ymm2       # ymm2 = ymm2[2,3,0,1]
        vmovdqu %ymm2, (%rdi,%rdx,2)
        vpshufb %ymm1, %ymm3, %ymm2
        vpermq  $78, %ymm2, %ymm2       # ymm2 = ymm2[2,3,0,1]
        vmovdqu %ymm2, -32(%rdi,%rdx,2)
        vpshufb %ymm1, %ymm4, %ymm2
        vpermq  $78, %ymm2, %ymm2       # ymm2 = ymm2[2,3,0,1]
        vmovdqu %ymm2, -64(%rdi,%rdx,2)
        vpshufb %ymm1, %ymm5, %ymm2
        vpermq  $78, %ymm2, %ymm2       # ymm2 = ymm2[2,3,0,1]
        vmovdqu %ymm2, -96(%rdi,%rdx,2)
        subq    $-128, %rsi
        addq    $-64, %rdx
        cmpq    %rdx, %r12
        jne     L304
; └└
```
so there is some vectorization in the loop happening, but I wouldn't do it now.

The same part with `enumerate` looks like this (on my cpu):
```asm
; │┌ @ int.jl:51 within `-'
        subw    (%rdx), %si
; │└
; │┌ @ words.jl:69 within `setindex!' @ array.jl:826
        movq    (%rbx), %rdi
        movw    %si, -2(%rdi,%r8,2)
; │└
```

PR #7, opened 2020-05-25
Alphabet initialization + Tests

bprzybylski:
This pull requests covers tests for the `Alphabet` struct and a new feature of creating an alphabet based on a predefined vector of symbols.


kalmarek:
btw. shall we keep the main repo clean, i.e. delete branches as we merge?

bprzybylski:
Actually we can. I had a thought that branches should be kept until a particular, independent functionality (e.g. alphabets) is implemented and documented in a consistent form (I omit the word "final" purposely). I agree that it would be more clean to create them each time so we avoid merging master into other branches.

As I think of it now, I do not know a good answer to your question. There are some pros and cons, so just make an arbitrary decision and we will stick to that.

kalmarek:
Ok, since we like to create small branches with small features lets delete branches from the main repo as they are merged. We can always push the branch from local if needed afterwards.

PR #8, opened 2020-05-25
Increase codecov for `alphabets.jl`

bprzybylski:
This pull request increases the code coverage of `alphabets.jl` file. However, it remains an open question whether it has any sense to test the output of the `Base.show` function.

(Yes, I am a small-stepper. Many small pull requests.)


kalmarek:
oops, I actually meant I'm `NOT buying` :P

PR #10, opened 2020-05-26
Orderings

mikolajpabiszczak:
Working LenLex ordering.

### DIFF
--- a/src/Orderings.jl
+++ b/src/Orderings.jl
@@ -1,0 +1,37 @@
+module Orderings
+
+import Base.Order: lt, Ordering
+export LexicographicOrdering, LexicographicOrder, LenLexOrder
+
+abstract type LexicographicOrdering <: Ordering end
+
+
+## Assuming the order is given by dictionary with entries (::"letter", ::int)
+struct LexicographicOrder <: LexicographicOrdering
+    order::Dict
+end
+
+
+struct LenLexOrder{LexOrd<:LexicographicOrdering} <: Ordering
+    lexord::LexOrd
+end
+
+# `T` shall be later changed to the type of letters of the Alphabet (Char?)
+lt(o::LexicographicOrder, a::T, b::T) = isless(o.order[a], o.order[b])
+
+
+## Assuming the following methods are implemented for type Word:
+## length, iteration
+## Assuming that for "letters" from Alphabet exists method: ==
+function lt(o::LenLexOrder, p::T, q::T) where T<:AbstractWord
+    if length(p) == length(q)
+        for i in 1:length(p)
+            p[i]==q[i] ||  return lt(o.lexord, p[i], q[i])
+        end
+        return false
+    else
+        return isless(length(p), length(q))
+    end
+end
+
+
--- a/src/Orderings.jl
+++ b/src/Orderings.jl
@@ -1,37 +1,29 @@
-module Orderings
-
 import Base.Order: lt, Ordering
-export LexicographicOrdering, LexicographicOrder, LenLexOrder
+export LenLexOrder
 
-abstract type LexicographicOrdering <: Ordering end
+"""
+    struct LenLexOrder{T<:Alphabet} <: Ordering
 
-
-## Assuming the order is given by dictionary with entries (::"letter", ::int)
-struct LexicographicOrder <: LexicographicOrdering
-    order::Dict
+Basic structure representing Length+Lexicographic (left-to-right) ordering of
+the words over given Alphabet. Lexicographing ordering of an Alphabet is
+implicitly specified inside Alphabet struct.
+"""
+struct LenLexOrder{T<:Alphabet} <: Ordering
+    A::T
 end
 
+"""
+    lt(o::LenLexOrder, p::T, q::T) where T<:AbstractWord
 
-struct LenLexOrder{LexOrd<:LexicographicOrdering} <: Ordering
-    lexord::LexOrd
-end
-
-# `T` shall be later changed to the type of letters of the Alphabet (Char?)
-lt(o::LexicographicOrder, a::T, b::T) = isless(o.order[a], o.order[b])
-
-
-## Assuming the following methods are implemented for type Word:
-## length, iteration
-## Assuming that for "letters" from Alphabet exists method: ==
+Return whether the first word is less then the other one in a given LenLex ordering.
+"""
 function lt(o::LenLexOrder, p::T, q::T) where T<:AbstractWord
     if length(p) == length(q)
-        for i in 1:length(p)
-            p[i]==q[i] ||  return lt(o.lexord, p[i], q[i])
+        for (a, b) in zip(p,q)
+            o.A[a] == o.A[b] || return isless(o.A[o.A[a]], o.A[o.A[b]])  # comparing only on positive pointer values
         end
         return false
     else
         return isless(length(p), length(q))
     end
 end
-
-
--- a/src/Orderings.jl
+++ b/src/Orderings.jl
@@ -1,26 +1,27 @@
 import Base.Order: lt, Ordering
-export LenLexOrder
+export LenLex
 
 """
-    struct LenLexOrder{T<:Alphabet} <: Ordering
+    struct LenLex{T} <: Ordering
 
 Basic structure representing Length+Lexicographic (left-to-right) ordering of
 the words over given Alphabet. Lexicographing ordering of an Alphabet is
 implicitly specified inside Alphabet struct.
 """
-struct LenLexOrder{T<:Alphabet} <: Ordering
-    A::T
+struct LenLex{T} <: Ordering
+    A::Alphabet{T}
 end
 
+
 """
-    lt(o::LenLexOrder, p::T, q::T) where T<:AbstractWord
+    lt(o::LenLex, p::T, q::T) where T<:Word{Integer}
 
 Return whether the first word is less then the other one in a given LenLex ordering.
 """
-function lt(o::LenLexOrder, p::T, q::T) where T<:AbstractWord
+function lt(o::LenLex, p::T, q::T) where T<:AbstractWord{<:Integer}
     if length(p) == length(q)
         for (a, b) in zip(p,q)
-            o.A[a] == o.A[b] || return isless(o.A[o.A[a]], o.A[o.A[b]])  # comparing only on positive pointer values
+            o.A[o.A[a]] == o.A[o.A[b]] || return isless(o.A[o.A[a]], o.A[o.A[b]])  # comparing only on positive pointer values
         end
         return false
     else
--- a/test/orderings.jl
+++ b/test/orderings.jl
@@ -1,13 +1,13 @@
 @testset "Orderings" begin
 
     import KnuthBendix.Alphabet, KnuthBendix.set_inversion!, KnuthBendix.Word
-    import KnuthBendix.LenLexOrder, Base.Order.lt
+    import KnuthBendix.LenLex, Base.Order.lt
 
-    A = Alphabet{String}(["a", "b", "c", "d"])
-    set_inversion!(A, "a", "b")
-    set_inversion!(A, "c", "d")
+    A = Alphabet(['a', 'b', 'c', 'd'])
+    set_inversion!(A, 'a', 'b')
+    set_inversion!(A, 'c', 'd')
 
-    lenlexord = LenLexOrder(A)
+    lenlexord = LenLex(A)
 
     @test lenlexord isa Base.Order.Ordering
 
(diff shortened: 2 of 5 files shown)


src/Orderings.jl
```diff
@@ -1,37 +1,29 @@
...
 
-struct LenLexOrder{LexOrd<:LexicographicOrdering} <: Ordering
-    lexord::LexOrd
+Basic structure representing Length+Lexicographic (left-to-right) ordering of
+the words over given Alphabet. Lexicographing ordering of an Alphabet is
+implicitly specified inside Alphabet struct.
+"""
+struct LenLexOrder{T<:Alphabet} <: Ordering
+    A::T
 end
```
bprzybylski: Is there any reason to consider the subtypes of `Alphabet`?
```julia
struct LenLexOrder <: Ordering
    A::Alphabet
end
```

src/Orderings.jl
```diff
@@ -1,37 +1,29 @@
...
-        for i in 1:length(p)
-            p[i]==q[i] ||  return lt(o.lexord, p[i], q[i])
+        for (a, b) in zip(p,q)
+            o.A[a] == o.A[b] || return isless(o.A[o.A[a]], o.A[o.A[b]])  # comparing only on positive pointer values
         end
         return false
     else
         return isless(length(p), length(q))
     end
 end
```
bprzybylski: For me, there are a few details omitted here.

- The assumption that `T <: AbstractWord` is too weak for the proposed implementation. How is this function going to behave for `Word{String}`? In line 23 you would call the alphabet using a string as the index which may not be valid if `A` is not the alphabet of strings. We need to get rid of either the generality, or the index calls.
- In line 23 you seek for a symbol in the alphabet, which is unnecessary. The symbols in the alphabet are equal if and only if `a` and `b` are equal. For the same reason, the call to `isless` is overcomplicated, because `A[A[x]] = x` if `x` is a valid index (or symbol). So, what I would suggest is
```julia
function lt(o::LenLexOrder, p::T, q::T) where T <: AbstractWord
    if length(p) == length(q)
        for (a, b) in zip(p,q)
            a == b || return isless(a, b) 
        end
        return false
    else
        return isless(length(p), length(q))
    end
end
```
but then, why do we even care what the alphabet is if we use numeric indices?

test/orderings.jl
```diff
@@ -0,0 +1,26 @@
+@testset "Orderings" begin
+
+    import KnuthBendix.Alphabet, KnuthBendix.set_inversion!, KnuthBendix.Word
+    import KnuthBendix.LenLexOrder, Base.Order.lt
+
+    A = Alphabet{String}(["a", "b", "c", "d"])
```
bprzybylski, line 6: Based on a sibling pull request it will be enough to write
```julia
A = Alphabet(["a", "b", "c", "d"])
```
However, maybe it is enough to create an alphabet of chars? I would need to update it in my tests.
```julia
A = Alphabet(['a', 'b', 'c', 'd'])
```

src/Orderings.jl
kalmarek: `Alphabet` is an abstract type, `Alphabet{T}` is concrete; we want concrete types in structs.

I'd probably go with
```julia
struct LenLexOrder{T} <: Ordering
    A::Alphabet{T}
end
```

src/Orderings.jl
kalmarek: for `LenLex` we actually don't care about the particular Alphabet.

My intention of implementing two `Base.getindex(A::Alphabet, x)` was precisely to make `A[A[i]] == i` for every valid index.

src/Orderings.jl
```diff
@@ -0,0 +1,29 @@
...
+export LenLexOrder
+
+"""
+    struct LenLexOrder{T<:Alphabet} <: Ordering
+
+Basic structure representing Length+Lexicographic (left-to-right) ordering of
+the words over given Alphabet. Lexicographing ordering of an Alphabet is
+implicitly specified inside Alphabet struct.
+"""
+struct LenLexOrder{T<:Alphabet} <: Ordering
```
kalmarek, line 11: maybe we don't even need to include `Order` in the name? what do you think?

src/Orderings.jl
mikolajpabiszczak: I believe we now run into certain problems with `Word` struct.

- I believe assumption is not too weak. What is too weak is the line 23 which should be of the form
`o.A[o.A[a]] == o.A[o.A[b]] || return isless(o.A[o.A[a]], o.A[o.A[b]])`
so that we will always compare pointers (they - i.e. positive pointers - store the implicit lexicographic order and iteration in `Word` is an iteration over the array of pointers).
- > the call to isless is overcomplicated, because `A[A[x]] = x` if `x`
Now that is not true:
```
A = KnuthBendix.Alphabet(["a", "b", "c", "d"])
KnuthBendix.set_inversion!(A, "a", "b")

julia> A[A[-1]]
2
```
and this is precisely why I used that weird construction (not seeing anything more sensible): `-1 == 2` will not work in the desired way (`isless` too). Which is - incidentally - a problem with the `==` method for words. I just checked and
```
u1 = KnuthBendix.Word([1,2])
u2 = KnuthBendix.Word([-2,-1])

julia> u1==u2
false
```
and should be `true` (the alphabet as above).

- > but then, why do we even care what the alphabet is if we use numeric indices?

I believe its role is to combine introduction of lexicographic order, inverses and - in a future - pretty printing of words.

test/orderings.jl
mikolajpabiszczak: Certainly, it will be enough. I just copied that line from the test and modified slightly (not reflecting too much on that).

src/Orderings.jl
bprzybylski: Okay, now I see why you used such an approach and it makes sense. But then, we need to assume that `T <: Word{Integer}`. Otherwise, the function will not work properly in general.

kalmarek: very well, you're correct! ;)
I had the simplistic view of the unique representation of a letter as a pointer, but that is incorrect.

So now we have a situation where words can be compared only in the context of an Alphabet.
hmm... that is less than ideal, but maybe I need to wrap my head around it;

as of `o.A[o.A[a]]` it does what it should (given that letters of Alphabet aren't integers), it's just not what I assumed it will :D

kalmarek: maybe my "smart" idea of using negative numbers as pointers to inverses wasn't that smart after all: without it we have a unique representation of each word...

src/Orderings.jl
mikolajpabiszczak: For this kind of questions I can just wave with 🇨🇭 flag and say I remain neutral. De gustibus non est disputandum.

src/Orderings.jl
mikolajpabiszczak: > But then, we need to assume that `T <: Word{Integer}`. Otherwise, the function will not work properly in general.

Yes, I see it now.

mikolajpabiszczak: I checked it, actually like that:
`function lt(o::LenLex, p::T, q::T) where T<:AbstractWord{<:Integer}`

kalmarek:
ok, so I've thought for a moment about negative numbers as inverses (which screws unique representations) and I don't think it's worth doing.
I.e. the uniqueness of word representations is more important than some "neat" functional overload.

So what I really would like to have is a way of quickly replacing a letter by its inverse (and by extension a quick inversion), while keeping the uniqueness of representation. Let's brainstorm, what are other options for achieving this.

* One way is to make an assumption that all entries in a word are positive (e.g. by making the internal storage to be `<:Unsigned`) and defining `invof(l::Integer, A::Alphabet{<:Integer})` that does either the double index, or would use `A.inversions`.

Share your ideas

src/Orderings.jl
kalmarek: > > But then, we need to assume that `T <: Word{Integer}`. Otherwise, the function will not work properly in general.
> 
> Yes, I see it now.

This was my silent assumption as well, so we can either enforce it here (as you did), or make 
```julia
abstract type AbstractWord{T<:Integer} <: AbstractVector{T} end
```

bprzybylski:
Here are some of my ideas and thoughts:

- Negative indices in the alphabet implementation of `getindex` seem to be a good choice. So I would not change that.
- Maybe we should introduce a new function, let's say `normalize`, that converts a word of integers to an equivalent word of unsigned ints. This way, we get rid of negative indices and can easily compare obtained words. Actually, I cannot see any other way to solve the problem.
- On the other hand, if we accept unnormalized words, we still have a possibility to safely remove a subsequence of, let's say, `[-6, 6]` without a single call to the alphabet, because we know it is a neutral element. I do not know how useful that is, but that is what I think of now.

mikolajpabiszczak:
I believe I have no other ideas. Nevertheless I'd say this:

- I do agree that the uniqueness of representation is more important.
- I am not sure if it is possible to have inversions "quick".

As I see it: the question is what will cost us more in the long run? From the point of view of KBS(1/2):
- we will do plenty of order comparisons (e.g. invoking `lt` for two right sides of rewriting rules with same left side or to keep left sides of rewriting rules in ascending order) - thus the less we will lose here, the better. So the suggested direction would be to get rid of that 'weird' 
```o.A[o.A[a]] == o.A[o.A[b]] || return isless(o.A[o.A[a]], o.A[o.A[b]])```
line from `lt` and have the simple `a == b || return isless(a, b)` thus the uniqueness of the representation via UInt would be beneficial;
- we will do a lot of searching for prefixes etc. and uniqueness of representation is crucial here (or else we will have to keep larger number of rules to cover the lack of uniqueness; what is even worse, we will have rules of which left or right sides will be equal with respect to order - this may have "serious repercussions");
- from the point of view of rewriting system (at least KBS(1/2)) inverses are not really important as those are declared via the rewriting rules (ab -> 1, ba -> 1 etc.); in that context their use would be more UI: easier declaring of words by users, possibly automatic addition of the rewriting rules for inverses.

That said I believe having uniqueness done via only positive entries in the `Word.ptrs` is sensible. We could have kind of the mentioned `invof(l::Integer, A::Alphabet{<:Integer})` and use it in conjunction with some additional user-friendly constructors like: `Word([1,2,-2,3])` that would work as `Word([1,2,invof(2),3])`. Similarly: one can use the same mechanism of inverting the word as we now have and just additionally calling this constructor at the and of "inverting" the array od ptrs.

I will stress again: this will not be of any use inside the rewriting system -
at the very moment I see no stage at which the rewriting system would generate a new word of the kind `Word([1,2,-2,3])`.

**Comment by kalmarek** [comment]
ok, I just tried the two approaches and it seems that it's hard to have consistent way of keeping inverses as negative numbers keep Words *decoupled* from alphabet;

The problem with the example
```julia
A = KnuthBendix.Alphabet(["a", "b", "c", "d"])
KnuthBendix.set_inversion!(A, "a", "b")

u1 = KnuthBendix.Word([1,2])
u2 = KnuthBendix.Word([-2,-1])

u1==u2 # false
```
is that these words are different, they only become equal when interpreted in the given Alphabet, but `==` has no idea about the alphabet, so `false` is the only sensible answer;

----
we could implement `isequal(w::Word, v:Word, A::Alphabet)` to return true in this case, but it will lead to lots of subtle errors when somebody forgets this and uses `==` to compare words;

* Solution 1: incorporate `Ref{Alphabet{T}}` into each `Word`. Then every operation on words brings `Alphabet` into the context and we're fine; `normalize!(w::Word)` which rewrites word to positive indices becomes an internal implementation detail/micro-optimization and can be implemented whenever is needed. (To avoid storing many copies of the same alphabet a reference only is stored.)
* Solution 2: keep the decoupling, and implement all operations on words with an additional argument `A::Alphabet`: we force the context of an alphabet onto words. `normalize!(w::Word, A::Alphabet)` can be implemented, but none of the algorithms can depend on uniqueness of words, unless it calls `normalize!` explicitly. Optimization here: words  store `isnormal::Bool` flag so that `normalize!` is no-op when not necessary. the flag will then be flipped only by direct modifications of the `ptrs` (e.g. `setindex!`, `push!`, `append` if the appended word is not normalized, etc), Con: `Word` becomes a `mutable struct` and we need to account for two options in most of the actions.
* Solution 3: we disallow negative entries in words (e.g. by `@assert`s in `setindex!`, `push!`, etc.); Words are now truly contexless, but  the possibility of quick (i.e. Alphabet-less) inversion of words is gone.

Instead of complicating things (Solution 1 and 2) I think we'll go for the third option now. It's better to keep the internals very simple for now, and we will probably return to the issue at later point (e.g. with Solution 1). Btw. Alphabets may keep their negative `getindex`, `Word`s will not use it.

PR #12, opened 2020-05-28
unique representations for words 

kalmarek:


src/alphabets.jl
```diff
@@ -213,3 +213,17 @@ function Base.getindex(A::Alphabet{T}, p::Integer) where T
 
     throw(DomainError("Inversion of $(A.alphabet[-p]) is not defined"))
 end
+
+"""
+    inv(w::Word, A::Alphabet)
+Return the inverse of a word `w` in the context of alphabet `A`.
+"""
+function Base.inv(w::AbstractWord, A::Alphabet)
```
bprzybylski, line 221: Docstring is not consistent with the declaration, I think.

src/words.jl
```diff
@@ -33,31 +33,37 @@ If type is not specified in the constructor it will default to `Int16`.
...
+    function Word{T}(v::AbstractVector{<:Integer}) where {T}
+        @assert all(x -> x > 0, v) "All entries of a Word must be positive integers"
+        return new{T}(v)
+    end
+    Word{T}(v::AbstractVector{<:Unsigned}) where {T<:Unsigned} = new{T}(v)
 end
 
 # setting the default type to Int16
-Word(x::Union{<:Vector{<:Integer}, <:AbstractVector{<:Integer}}) = Word{Int16}(x)
+Word(x::Union{<:Vector{<:Integer},<:AbstractVector{<:Integer}}) = Word{UInt16}(x)
```
bprzybylski, line 45: Well, I think that something is wrong with the requirement that all the elements of a word must be unsigned integers, but I cannot tell you exactly what it is. I need to think about that.

By the way, am I seeing it correctly that for a vector of integers it is not allowed for any pointer to be less or equal to 0, but in case of unsigned integers it is allowed for a pointer to be 0?

mikolajpabiszczak:
Just an idea concerning words: I think `pop!` and `popfirst!` should also be implemented for `Word` - will be useful in rewriting.

src/words.jl
kalmarek: yes, that was a premature optimization

src/alphabets.jl
kalmarek: thanks, done!

kalmarek:
> Just an idea concerning words: I think `pop!` and `popfirst!` should also be implemented for `Word` - will be useful in rewriting.

implement at will, when needed

test/words.jl
```diff
@@ -1,51 +1,50 @@
...
-    
-    @test Word([1,2])*Word([2,3]) == Word([1,2,2,3])
-    @test Word([1,2])*W == Word([1,2,1,-2])
-    
-    @test inv(W) == Word([2,-1])
-    @test inv(Word([1,-2,2])) == Word([-2, 2, -1])
-    @test inv(inv(w)) == w
+
+    @test Word([1, 2]) * Word([2, 3]) == Word([1, 2, 2, 3])
+    @test Word([1, 2]) * W == Word([1, 2, 1, 2])
```
mikolajpabiszczak, line 49: `inv` is back but tests for it are not.

kalmarek: `inv` needs now an alphabet, so those tests were moved to `test/alphabets.jl`

PR #13, opened 2020-05-31
Working kbs1, lenlexordering and minor updates

mikolajpabiszczak:
Working `kbs1` together with `LenLexOrdering` that now takes into account that words consist only of positive pointers. Also added some additional methods for `Word`. 

I checked with the examples from Sims: reproduced tables 2.5.1, 2.5.3, 2.5.4 (in the last one there is an error in the book: rule 13 shall not be in the table) and also reproduced the sets of final systems.

Note: I did not try to optimise the code (there was no point for `kbs1`), it is basically 1-1 what was in the book.

Remark: I did `AbstractRewritingSystem` and `RewritingSystem` similarly to the way `Word` was done. I am not sure it is good though.

To consider for later: right now when we iterate over `Word` we obtain Integers. If we subset `Word` via iterator we obtain a vector. Maybe it is sensible to always obtain `Word` via that protocol, i.e. `Word([1,2,3])[2] = Word([2])` and `Word([1,2,3])[2:3] = Word([2,3])`. It would make the code for `kbs1` clearer (shall I open an issue to discuss it separately? This is not something important and maybe later it will become more evident in which way this should be solved).

### DIFF
--- a/src/kbs1.jl
+++ b/src/kbs1.jl
@@ -1,10 +1,10 @@
 """
-    test1!(u::Word, v::Word, rs::RewritingSystem, o::Ordering)
+    test1!(rs::RewritingSystem, u::Word, v::Word, o::Ordering)
 Adds a rule to a rewriting system (if necessary) that insures that there is
 a word derivable form two given words using the rules in rewriting system.
 See [Sims, p. 69].
 """
-function test1!(u::Word, v::Word, rs::RewritingSystem, o::Ordering)
+function test1!(rs::RewritingSystem, u::Word, v::Word, o::Ordering)
     a = rewrite_from_left(u, rs)
     b = rewrite_from_left(v, rs)
     if a != b
@@ -13,18 +13,20 @@ function test1!(u::Word, v::Word, rs::RewritingSystem, o::Ordering)
 end
 
 """
-    overlap1!(i::Integer, j::Integer, rs::RewritingSystem, o::Ordering)
+    overlap1!(rs::RewritingSystem, i::Integer, j::Integer, o::Ordering)
 Checks the overlaps of right sides of rules at position i and j in the rewriting
-system in which rule at i occurs at the beginning of the word. Adds new rules
-if failures of local confluence are found. See [Sims, p. 69].
+system in which rule at i occurs at the beginning of the word. When failures of
+local confluence are found, new rules are added. See [Sims, p. 69].
 """
-function overlap1!(i::Integer, j::Integer, rs::RewritingSystem, o::Ordering)
-    for k in 1:length(rs[i].first)
-        a = Word(rs[i].first[1:end-k])
-        b = Word(rs[i].first[end-k+1:end])
-        n = lcp(b, rs[j].first)
-        if isone(Word(b[n+1:end])) || isone(Word(rs[j].first[n+1:end]))
-            test1!(a * rs[j].second * Word(b[n+1:end]), rs[i].second * Word(rs[j].first[n+1:end]), rs, o)
+function overlap1!(rs::RewritingSystem,i::Integer, j::Integer,  o::Ordering)
+    lhs_i, rhs_i = rules(rs)[i]
+    lhs_j, rhs_j = rules(rs)[j]
+    for k in 1:length(lhs_i)
+        a = Word(lhs_i[1:end-k])
+        b = Word(lhs_i[end-k+1:end])
+        n = lcp(b, lhs_j)
+        if isone(Word(b[n+1:end])) || isone(Word(lhs_j[n+1:end]))
+            test1!(rs, a * rhs_j * Word(b[n+1:end]), rhs_i * Word(lhs_j[n+1:end]), o)
         end
     end
 end
@@ -36,47 +38,47 @@ proper subwords are irreducible with respect to this rewriting system.
 """
 function getirrsubsys(rs::RewritingSystem)
     rsides = []
-    for rule in rs
+    for (lhs, rhs) in rules(rs)
         ok = true
-        n = length(rule.first)
+        n = length(lhs)
         if n > 2
             for j in 2:(n-1)
-                w = Word(rule.first[1:j])
+                w = Word(lhs[1:j])
                 rw = rewrite_from_left(w, rs)
                 (w == rw) || (ok=false; break)
             end
             for i in 2:(n-1)
                 ok || break
                 for j in (i+1):n
-                    w = Word(rule.first[i:j])
+                    w = Word(lhs[i:j])
                     rw = rewrite_from_left(w, rs)
                     (w == rw) || (ok=false; break)
                 end
             end
         end
-        ok && push!(rsides, rule.first)
+        ok && push!(rsides, lhs)
     end
     return rsides
 end
 
 """
-    kbs1(rs::RewritingSystem, o::Ordering)
+    knuthbendix1(rs::RewritingSystem, o::Ordering)
 Implements a Knuth-Bendix algorithm that yields reduced, confluent rewriting
 system. See [Sims, p.68].
 
 Warning: termination may not occur.
 """
-function kbs1(rs::RewritingSystem, o::Ordering)
+function knuthbendix1(rs::RewritingSystem, o::Ordering)
     i = 1
     ss = zero(rs)
 
-    for rule in rs
-        test1!(rule.first, rule.second, ss, o)
+    for (rhs, lhs) in rules(rs)
+        test1!(ss, lhs, rhs, o)
     end
     while i ≤ length(ss)
         for j in 1:i
-            overlap1!(i,j, ss, o)
-            j<i && overlap1!(j,i, ss, o)
+            overlap1!(ss, i,j, o)
+            j<i && overlap1!(ss, j,i, o)
         end
         i += 1
     end
--- a/src/kbs1.jl
+++ b/src/kbs1.jl
@@ -4,7 +4,7 @@ Adds a rule to a rewriting system (if necessary) that insures that there is
 a word derivable form two given words using the rules in rewriting system.
 See [Sims, p. 69].
 """
-function test1!(rs::RewritingSystem, u::Word, v::Word, o::Ordering)
+function test1!(rs::RewritingSystem{W,O}, u::W, v::W, o::Ordering = ordering(rs))  where {W<:AbstractWord, O<:Ordering}
     a = rewrite_from_left(u, rs)
     b = rewrite_from_left(v, rs)
     if a != b
@@ -18,15 +18,15 @@ Checks the overlaps of right sides of rules at position i and j in the rewriting
 system in which rule at i occurs at the beginning of the overlap. When failures
 of local confluence are found, new rules are added. See [Sims, p. 69].
 """
-function overlap1!(rs::RewritingSystem,i::Integer, j::Integer,  o::Ordering)
+function overlap1!(rs::RewritingSystem{W,O},i::Integer, j::Integer,  o::Ordering = ordering(rs)) where {W<:AbstractWord, O<:Ordering}
     lhs_i, rhs_i = rules(rs)[i]
     lhs_j, rhs_j = rules(rs)[j]
     for k in 1:length(lhs_i)
-        a = Word(lhs_i[1:end-k])
-        b = Word(lhs_i[end-k+1:end])
+        a = W(lhs_i[1:end-k])
+        b = W(lhs_i[end-k+1:end])
         n = longestcommonprefix(b, lhs_j)
-        if isone(Word(b[n+1:end])) || isone(Word(lhs_j[n+1:end]))
-            test1!(rs, a * rhs_j * Word(b[n+1:end]), rhs_i * Word(lhs_j[n+1:end]), o)
+        if isone(W(b[n+1:end])) || isone(W(lhs_j[n+1:end]))
+            test1!(rs, a * rhs_j * W(b[n+1:end]), rhs_i * W(lhs_j[n+1:end]), o)
         end
     end
 end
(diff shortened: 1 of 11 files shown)


kalmarek:
that's great!

> To consider for later: right now when we iterate over Word we obtain Integers. If we subset Word via iterator we obtain a vector. Maybe it is sensible to always obtain Word via that protocol, i.e. Word([1,2,3])[2] = Word([2]) and Word([1,2,3])[2:3] = Word([2,3]). It would make the code for kbs1 clearer (shall I open an issue to discuss it separately?

Please open an issue to discuss this

src/words.jl
```diff
@@ -74,3 +80,18 @@ Base.@propagate_inbounds function Base.setindex!(w::Word, v::Integer, n::Integer
...
+    lcp(u::Word, v::Word)
+Returns the length of longest common prefix of two words (and simultaneously
+the index at which the prefix ends).
+"""
+function lcp(u::Word, v::Word)
+    n=0
+    for i in zip(u,v)
+        if i[1] == i[2]
+            n += 1
+        end
```
kalmarek, line 94: if we're using both elts in zip, then it's better to
```julia
for (lu, lv) in zip(u,v)
    lu != lv && break
    n += 1
end
```
indexing to `i`has small penalty for bounds check, or maybe used to have?

src/words.jl
```diff
@@ -74,3 +80,18 @@ Base.@propagate_inbounds function Base.setindex!(w::Word, v::Integer, n::Integer
     @assert v > 0 "All entries of a Word must be positive integers"
     return @inbounds w.ptrs[n] = v
 end
+
+"""
+    lcp(u::Word, v::Word)
+Returns the length of longest common prefix of two words (and simultaneously
+the index at which the prefix ends).
+"""
+function lcp(u::Word, v::Word)
```
kalmarek, line 89: I'd prefer `longestprefix` or something more wordy

src/rewriting.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+    rewrite_from_left(u::Word, rs::RewritingSystem)
+Rewrites a word from left using rules from a given RewritingSystem. See [Sims, p.66]
+"""
+function rewrite_from_left(u::Word, rs::RewritingSystem)
+    iszero(rs) && return u
+    v = one(u)
+    w = copy(u)
+    while !isone(w)
+        v *= Word(popfirst!(w))
```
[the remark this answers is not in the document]
kalmarek: probably we should implement `sizehint!(w::Word)` and hint the size of `v` to avoid repetetive allocations (`length(u)` or maybe half of it?, this needs some heuristics)

src/rewriting.jl
```diff
@@ -0,0 +1,82 @@
...
+`AbstractRewritingSystem` as such has its meaning only in the contex of an Alphabet.
+The subtypes of `AbstractRewritingSystem{T}` need to implement the following
+methods which constitute `AbstractRewritingSystem` interface:
+ * `Base.push!`/`Base.pushfirst!`: appending a single rule at the end/beginning
+ * `Base.append!`/`Base.prepend!`: appending a another system at the end/beginning,
+ * `Base.insert!`: inserting a single rule at a given place
+ * `Base.delateat!`: delating rules at given positions
+ * `Base.empty!`: delating all the rules
+ * full iteration protocol for `AbstractArray`s, returning pairs of Words (rules)
+ * `length` the length of word as written in the alphabet.
```
kalmarek, line 14: copy-paste errors for these two lines

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
...
+
+Warning: termination may not occur.
+"""
+function kbs1(rs::RewritingSystem, o::Ordering)
+    i = 1
+    ss = zero(rs)
+
+    for rule in rs
+        test1!(rule.first, rule.second, ss, o)
+    end
```
kalmarek, line 75:
```suggestion
    for (rhs, lhs) in rules(rs)
        test1!(lhs, rhs, ss, o)
    end
```

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
+"""
+    test1!(u::Word, v::Word, rs::RewritingSystem, o::Ordering)
+Adds a rule to a rewriting system (if necessary) that insures that there is
+a word derivable form two given words using the rules in rewriting system.
+See [Sims, p. 69].
+"""
+function test1!(u::Word, v::Word, rs::RewritingSystem, o::Ordering)
```
kalmarek, line 7: the general practice is to place the first argument as the one being modified, i.e.
```suggestion
function test1!(rs::RewritingSystem, u::Word, v::Word, o::Ordering = ordering(rs))
```
I also added the default ordering if we include it into the rewriting system

We open a contest for better name for this function :D my candidate is `deriverule!`

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
...
+    a = rewrite_from_left(u, rs)
+    b = rewrite_from_left(v, rs)
+    if a != b
+        lt(o, a, b) ? push!(rs, b=>a) : push!(rs, a=>b)
+    end
+end
+
+"""
+    overlap1!(i::Integer, j::Integer, rs::RewritingSystem, o::Ordering)
```
kalmarek, line 16: same comment about the order of arguments.
as a name I suggest `isconfluent_local` (there will be `isconfluent(rws::RewritingSystem)`)

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
...
+Checks the overlaps of right sides of rules at position i and j in the rewriting
+system in which rule at i occurs at the beginning of the word. Adds new rules
```
kalmarek, line 18: the second part of the first sentence is not clear to me

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+
+"""
+    kbs1(rs::RewritingSystem, o::Ordering)
+Implements a Knuth-Bendix algorithm that yields reduced, confluent rewriting
+system. See [Sims, p.68].
+
+Warning: termination may not occur.
+"""
+function kbs1(rs::RewritingSystem, o::Ordering)
```
kalmarek, line 69:
```suggestion
function knuthbendix1(rs::RewritingSystem, o::Ordering=ordering(rs))
```

src/kbs1.jl
```diff
@@ -0,0 +1,91 @@
...
+function overlap1!(i::Integer, j::Integer, rs::RewritingSystem, o::Ordering)
+    for k in 1:length(rs[i].first)
+        a = Word(rs[i].first[1:end-k])
+        b = Word(rs[i].first[end-k+1:end])
+        n = lcp(b, rs[j].first)
+        if isone(Word(b[n+1:end])) || isone(Word(rs[j].first[n+1:end]))
+            test1!(a * rs[j].second * Word(b[n+1:end]), rs[i].second * Word(rs[j].first[n+1:end]), rs, o)
+        end
+    end
+end
```
kalmarek, line 30: there is lots of indexing here, how about using
```julia
lhs_i, rhs_i = rules(rs)[i]
lhs_j, rhs_j = rules(rs)[j]
```

kalmarek:
ok, maybe I went overboard with the number of comments, but the gist is that I like what I see ;)

Besides indexing into words, that needs special care, there is one thing that we need to be careful about: the code above uses the default `Word` constructor, but we should rather respect what the user gave us. E.g. stating `rws::RewritingSystem{W where W<:AbstractWord}` in the signature and later using `W(...)` instead of `Word(...)`. Most of this should be fixed by proper (range) indexing into `Word`s.

src/words.jl
mikolajpabiszczak: Yes and no. `lcp` seems to be well known: cf. [LCP array](https://en.wikipedia.org/wiki/LCP_array).

src/rewriting.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+
+abstract type AbstractRewritingSystem{T} <: AbstractVector{T} end
+
+"""
+    RewritingSystem{T} <: AbstractRewritingSystem{T}
+RewritingSystem written as a list of pairs of Words (left side => right side)
+"""
+struct RewritingSystem{T} <: AbstractRewritingSystem{T}
+    rwrules::Vector{T}
```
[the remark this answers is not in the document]
mikolajpabiszczak: This has pros and cons:
- in general `RewritingSystem` needs no `Ordering` - it is just a set of rules (I think of it as the "~" part  of "< ... |  ~  >"  monoid/group);
- as to `test1!` and `overlap1!` - at that stage I wanted to be consistent with Sims' book. After all I believe that the real thing will start with `kbs2` (it has substantial differences) - names are just a final polish. Keeping things consistent with the book has some advantage at that stage: it allows to easily look up things - seems trivial, but ...
- `getirrsubsys` - this is probably only important in `kbs1`, later there will be no such thing as we will remove rules at the earlier stage ('active'/'inactive' in the book). BTW: starting from `kbs2` it may be useful to make `RewritingSystem` possess a dictionary of rules instead of a list of rules (as the left sides will be always unique according to the book).

Thus answering the first (implicit) question: I would opt for not having too much in `RewritingSystem` at the moment - it will make things more rigid and I am not sure it is a stage at which we want things to be too rigid.

src/rewriting.jl
mikolajpabiszczak: That would be better:
```
 * full iteration protocol for `AbstractRewritingSystem`s, returning pairs of Words (rules)
 * `length`: the number of rules (not necessarily unique) stored inside the system
```
?

src/kbs1.jl
mikolajpabiszczak: `When failures of local confluence are found, new rules are added.` would be better? (verbatim from Sims)

mikolajpabiszczak:
As to the names of functions: I only changed `kbs1` to `knuthbendix1`. The rest is not worth doing now. We can use the new naming starting from `kbs2`. I will open new issue to gather the propositions that appeared so far.

src/words.jl
mikolajpabiszczak: My personal preference:
`lcp` > `longestcommonprefix` > `longestprefix`
so just tell me which one. 😁

src/kbs1.jl
kalmarek: i meant the first sentence: `Checks the overlaps of right sides of rules at position i and j in the rewriting system in which rule at i occurs at the beginning of the word`
what is `the word` here?

src/words.jl
kalmarek: then `longestcommonprefix` let it be

src/rewriting.jl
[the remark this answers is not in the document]
kalmarek: that is true, but only if we always start from empty `RewritingSystem`. Otherwise (e.g. resuming `knuthbendix` after increasing the number of rules, or just running again with some time-budget) we need to check if the existing rules in `rws` are consistent with the given ordering.

I'm fine with the names for now, this was just to spark some discussion;

the ordering can stay outside of `RewritingSystem` for now (before we dig into `kbs2` and automata), but I guess it will be much easier to reason about the functions (`knuthbendix` in particular) when ordering is stored inside and cannot be changed in the middle of lifespan of an `rws`

src/rewriting.jl
```diff
@@ -24,31 +24,31 @@ struct RewritingSystem{T} <: AbstractRewritingSystem{T}
...
+Base.iszero(s::RewritingSystem) = isempty(rules(s))
 
-Base.push!(s::RewritingSystem, r::Pair{T,T}) where {T<:AbstractWord} = (push!(s.rwrules, r); s)
-Base.pushfirst!(s::RewritingSystem, r::Pair{T,T}) where {T<:AbstractWord} = (pushfirst!(s.rwrules, r); s)
+Base.push!(s::RewritingSystem, r::Pair{T,T}) where {T<:AbstractWord} = (push!(rules(s), r); s)
+Base.pushfirst!(s::RewritingSystem, r::Pair{T,T}) where {T<:AbstractWord} = (pushfirst!(rules(s), r); s)
 
-Base.append!(s::RewritingSystem, v::RewritingSystem) = (append!(s.rwrules, v.rwrules); s)
-Base.prepend!(s::RewritingSystem, v::RewritingSystem) = (prepend!(s.rwrules, v.rwrules); s)
+Base.append!(s::RewritingSystem, v::RewritingSystem) = (append!(rules(s), v.rwrules); s)
```
kalmarek, line 40:
```suggestion
Base.append!(s::RewritingSystem, v::RewritingSystem) = (append!(rules(s), rules(v)); s)
```

kalmarek: similarly below

src/rewriting.jl
```diff
@@ -0,0 +1,82 @@
...
+
+Base.append!(s::RewritingSystem, v::RewritingSystem) = (append!(s.rwrules, v.rwrules); s)
+Base.prepend!(s::RewritingSystem, v::RewritingSystem) = (prepend!(s.rwrules, v.rwrules); s)
+
+Base.insert!(s::RewritingSystem, i::Integer, r::Pair{T,T}) where {T<:AbstractWord} = (insert!(s.rwrules, i, r); s)
+Base.deleteat!(s::RewritingSystem, i::Integer) = (deleteat!(s.rwrules, i); s)
+Base.deleteat!(s::RewritingSystem, inds) = (deleteat!(s.rwrules, inds); s)
+Base.empty!(s::RewritingSystem) = (empty!(s.rwrules); s)
+
+Base.iterate(s::RewritingSystem) = iterate(s.rwrules)
```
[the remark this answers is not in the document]
kalmarek: I don't think it has much value to allow direct iteration over `RewritingSystem` now, when iteration over `rules(rws)` describes the intent much clearer

src/kbs1.jl
mikolajpabiszczak: Ahh 😅. `the Word` is the same as `the overlap` from the beginning of the docstring.
```
Checks the overlaps of right sides of rules at position i and j in the rewriting
system in which rule at i occurs at the beginning of the overlap.
```
should be more clear.

src/rewriting.jl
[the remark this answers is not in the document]
mikolajpabiszczak: Yes, I am going to get rid of it; will do it today.

src/rewriting.jl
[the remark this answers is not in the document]
mikolajpabiszczak: OK, now I see the argument for including the `Ordering` into the `RewritingSystem` and will  implement it.

src/rewriting.jl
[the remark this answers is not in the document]
mikolajpabiszczak: Although there is a question whether methods `push!(s::RewritingSystem,...)`, `append!(...)` etc. shall be left (this looks a bit inconsistent with lack of iteration) or shall be removed and we just use calls of the form `push!(rules(s),..)` (which do not require implementation).

mikolajpabiszczak:
> [...] one thing that we need to be careful about: the code above uses the default `Word` constructor, but we should rather respect what the user gave us. E.g. stating `rws::RewritingSystem{W where W<:AbstractWord}` in the signature and later using `W(...)` instead of `Word(...)`. Most of this should be fixed by proper (range) indexing into `Word`s.

Done.

PR #16, opened 2020-06-04
indexing into Words and SubWord

kalmarek:
and some eye-candy

### DIFF
--- a/src/words.jl
+++ b/src/words.jl
@@ -41,41 +41,51 @@ struct Word{T} <: AbstractWord{T}
     end
 end
 
-Word(n::Integer) = Word([n])
-
 # setting the default type to Int16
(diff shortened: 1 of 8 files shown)


src/words.jl
```diff
@@ -86,11 +96,11 @@ end
 Returns the length of longest common prefix of two words (and simultaneously
 the index at which the prefix ends).
 """
-function longestcommonprefix(u::Word, v::Word)
+function longestcommonprefix(u::Union{Word, SubWord}, v::Union{Word, SubWord})
```
bprzybylski, line 99: Maybe it would be a good idea to call this function using an abbrev.? Maybe `lcp`/`LCP`? I know it is not that important, but I cannot stop imaging `greatestcommondivisor` and it makes me feel dizzy.

kalmarek: yeah, I might be leaning towards it as well. Despite the fact that I asked for the wordy name in the first place :laughing:

PR #17, opened 2020-06-13
Working kbs2

mikolajpabiszczak:
Pardon this very crude implementation.

I had to change `RewritingSystem` a bit. Decided not to totally reshape it yet (for instance I did not change `==` to do comparison on active rules only) - wasn't sure if we will stick to those active/inactive rules. Was thinking about removing them when they become inactive, but that messes with iteration...

Ofc. I will change naming of `test2!` and `overlap2!` with some later commit.


src/words.jl
```diff
@@ -104,3 +104,32 @@ function longestcommonprefix(u::Union{Word, SubWord}, v::Union{Word, SubWord})
     end
     return n
 end
+
+"""
+    issubword(u::Union{Word, SubWord}, v::Union{Word, SubWord})
```
kalmarek, line 109: how does this compare to
```julia
using KnuthBendix
using BenchmarkTools

import KnuthBendix: AbstractWord

function Base.findfirst(subword::AbstractWord, word::AbstractWord)
    k = length(subword)
    for i in 1:length(word)-k+1
        @view(word[i:i+k-1]) == subword && return i
    end
    return nothing
end

Base.occursin(subword::AbstractWord, word::AbstractWord) = findfirst(subword, word) !== nothing

let W = KnuthBendix.Word(1:100), s = KnuthBendix.Word([3,2,1]) #worst case scenario
    @btime occursin($s, $W) 
end
```
could you btime?
in case of single letters (words of length 1) we could add simpler loops;

(the above allocates views on julia-1.4, but these allocations go away with julia-1.5; I'm fine with this

kalmarek:
thanks! I will have a detailed look at this before Tuesday.

src/words.jl
mikolajpabiszczak: That's something of a kind "was looking for cooper and found gold", here are the results: 
- **occursin**
```
let W = KnuthBendix.Word(1:100), s = KnuthBendix.Word([3,2,1]) #worst case scenario
    @btime occursin($s, $W)
end
```
1.130 μs (98 allocations: 4.59 KiB)
false

- **issubword**
```
let W = KnuthBendix.Word(1:100), s = KnuthBendix.Word([3,2,1]) #worst case scenario
    @btime KnuthBendix.issubword($s, $W)
end
```
72.532 ns (0 allocations: 0 bytes)
false

src/words.jl
```diff
@@ -92,7 +94,7 @@ Base.@propagate_inbounds function Base.setindex!(w::Union{Word, SubWord}, v::Int
 end
 
 """
-    longestcommonprefix(u::Word, v::Word)
+    longestcommonprefix(u::Union{Word, SubWord}, v::Union{Word, SubWord})
```
kalmarek, line 97:
```suggestion
    longestcommonprefix(u::AbstractWord, v::AbstractWord)
```
whenever we don't use `ptrs` field the method should be applicable to `AbstractWord`s

src/words.jl
kalmarek: On my computer this:
```julia
function Base.findnext(subword::AbstractWord, word::AbstractWord, pos::Integer)
    k = length(subword)
    f = first(subword)
    @inbounds for i in pos:length(word)-k+1
        # @view(word[i:i+k-1]) == subword && return i
        word[i] == f || continue
        issub = true
        for j in 2:k
            if word[i+j-1] != subword[j]
                issub = false
                break
            end
        end
        issub == true && return i:i+k-1
    end
    return nothing
end

@inline Base.findfirst(subword::AbstractWord, word::AbstractWord) = findnext(subword, word, firstindex(word))
@inline Base.occursin(subword::AbstractWord, word::AbstractWord) = findfirst(subword, word) !== nothing
```
is a constant factor slower that `issubword` (95ns vs 88ns on my computer). This is actually something we want to have reasonable performance

src/kbs2.jl
```diff
@@ -0,0 +1,89 @@
...
+        a = rewrite_from_left(lr, rs)
+        b = rewrite_from_left(rr, rs)
+        if a != b
+            if lt(o, a, b)
+                a, b = b, a
+            end
+            push!(rs, a => b)
+
+            for (i, (lhs, rhs)) in enumerate(rules(rs)[1:end-1])
+                if active(rs)[i] && issubword(a, lhs)
```
kalmarek, line 19:
```suggestion
            for i in 1:length(rules(rs))-1
                (lhs, rhs) = rules(rs)[i]
                if isactive(rs, i) && issubword(a, lhs)
```

src/kbs2.jl
```diff
@@ -0,0 +1,89 @@
+"""
+    test2!(rs::RewritingSystem, u::Word, v::Word, o::Ordering)
```
kalmarek, line 2: update signature

src/kbs2.jl
```diff
@@ -0,0 +1,89 @@
...
+                else issubword(a, rhs)
+                    rules(rs)[i] = (lhs => rewrite_from_left(rhs, rs))
+                end
+            end
+        end
+    end
+end
+
+"""
+    overlap2!(rs::RewritingSystem, i::Integer, j::Integer, o::Ordering)
```
kalmarek, line 31: update signature

src/rewriting.jl
```diff
@@ -20,26 +21,39 @@ RewritingSystem written as a list of pairs of `Word`s together with the ordering
...
     rwrules::Vector{Pair{W,W}}
     order::O
+    act::BitArray{1}
 end
 
+RewritingSystem(rwrules::Vector{Pair{W,W}}, order::O) where{W<:AbstractWord, O<:Ordering} = RewritingSystem(rwrules, order, trues(length(rwrules)))
+
 rules(s::RewritingSystem) = s.rwrules
+active(s::RewritingSystem) = s.act
+arules(s::RewritingSystem) = rules(s)[active(s)]
```
kalmarek, line 31: this allocates a copy. If we just want to iterate, then a generator would do: `(r for (i,r) in enumerate(rules(s)) if isactive(s, i))`

src/rewriting.jl
```diff
@@ -20,26 +21,39 @@ RewritingSystem written as a list of pairs of `Word`s together with the ordering
...
 Base.hash(s::RewritingSystem, h::UInt) =
     foldl((h, x) -> hash(x, h), s.rwrules, init = hash(s.order, hash(0x905098c1dcf219bc, h)))
 # the init value is simply hash(RewritingSystem)
 
-Base.push!(s::RewritingSystem{W,O}, r::Pair{W,W}) where {W<:AbstractWord, O<:Ordering} = (push!(rules(s), r); s)
-Base.pushfirst!(s::RewritingSystem{W,O}, r::Pair{W,W}) where {W<:AbstractWord, O<:Ordering} = (pushfirst!(rules(s), r); s)
+Base.push!(s::RewritingSystem{W,O}, r::Pair{W,W}) where {W<:AbstractWord, O<:Ordering} = (push!(rules(s), r); push!(active(s), true); s)
+Base.pushfirst!(s::RewritingSystem{W,O}, r::Pair{W,W}) where {W<:AbstractWord, O<:Ordering} = (pushfirst!(rules(s), r); pushfirst!(active(s), true); s)
+
+Base.pop!(s::RewritingSystem) = (deleteat!(active(s), length(s)); pop!(rules(s)))
```
kalmarek, line 47: why not `Base.pop!(s::RewritingSystem) = (pop!(active(s)); pop!(rules(s)))`?

src/words.jl
```diff
@@ -104,3 +106,30 @@ function longestcommonprefix(u::Union{Word, SubWord}, v::Union{Word, SubWord})
     end
     return n
 end
+
```
kalmarek, line 109:
```suggestion
"""
    lcp(u::AbstractWord, v::AbstractWord)
See [`longestcommonprefix`](@ref).
"""
lcp(u::AbstractWord, v::AbstractWord) = longestcommonprefix(u,v)
```

src/rewriting.jl
mikolajpabiszczak: Good question. I have no idea what craziness took hold of me. Corrected.

src/words.jl
mikolajpabiszczak: Changing from `issubword` to `occursin` and purging `issubword` due to copyright discussion.

mikolajpabiszczak:
Few questions/issues:
- should `==` be now performed only on active rules or both active and inactive (I believe only on active)?
- How about `show` function: only showing active rules or both active/inactive (say with "active" or "inactive" next to it)?
- For `knuthbendix2`: shall I add another (optional) parameter `keepinactiverules` (as to discussion with @kalmarek to keep kbs1/kbs2 for educational purposes)?

kalmarek:
Thanks for all the changes;
* I'm going to dodge the first question by saying that I don't think we should even implement the equality :D. Is it mathematical equality? That should probably mean that all rules in one rws are derivable in the second and vice-versa. Which includes checking confluence and running Knuth-Bendix. This is way too long for practical purposes. Besides: do we actually need it? ;)
* marking inactive rules in `show` would be nice, but only very small rwses can be examined visually, anyway (we could elide the excessive output as well).
* let's postpone this for later

bprzybylski:
I just wanted to let you know that I read the code and I am now up to date.

src/kbs2.jl
```diff
@@ -68,7 +69,8 @@ function knuthbendix2!(rs::RewritingSystem, o::Ordering = ordering(rs))
     end
 
     i = 1
-    while i ≤ length(tmprs)
+    while i ≤ (ltmprs = length(tmprs))
+        ltmprs > maxrules && (println("Maximum number of rules in the RewritingSystem reached. You can try again with higher value."); break)
```
kalmarek, line 73: no `print`ing of messages these go to stdout, while user may be watching a logger; use `@warn(msg)`

mikolajpabiszczak:
Ad `==`: we do not really need it. In the very beginning I hesitantly decided to implement it just to test things with less writing (with the hope that somebody points to that as at least fairly stupid). So I think it is time to clean up that mess and get rid of it.

Issue #18, opened 2020-06-16
is getirrsubsystem correct?


kalmarek:
As implemented here:
https://github.com/kalmarek/KnuthBendix.jl/blob/342b15ef14e9ad1f0c437e94baea7308c08dd0f6/src/kbs1.jl#L43
shouldn't we test all subwords, not only prefixes and suffixes?

btw. rewriting and comparing a word is a poor test for irreducibility. This could be implemented e.g.
```julia
function isirreducible(w::AbstractWord, rws::RewritingSystem)
    for (lhs, _) in rules(rws)
        issubword(lhs, w) && lhs != w && return false
    end
    return true
end
```
without allocations

mikolajpabiszczak:
Aren't we testing all subwords? 

Here we first check prefixes:
https://github.com/kalmarek/KnuthBendix.jl/blob/342b15ef14e9ad1f0c437e94baea7308c08dd0f6/src/kbs1.jl#L49-L53

But here I check everything else (there are two for-loops):
https://github.com/kalmarek/KnuthBendix.jl/blob/342b15ef14e9ad1f0c437e94baea7308c08dd0f6/src/kbs1.jl#L54-L60

(Two separate loops in order to ensure that I check only proper subwords).

As to the rest of the issue: absolutely right, will fix that.

kalmarek:
very well, i must have missed the inner loop!

PR #20, opened 2020-06-30
More efficient getirrsubys

mikolajpabiszczak:
More efficient `getirrsubys`, docstring fix, more consistent variables naming (`getirrsubsys` corrected as suggested in #18).

### DIFF
--- a/src/kbs1.jl
+++ b/src/kbs1.jl
@@ -40,7 +40,7 @@ Returns whether a word is irreducible with respect to a given rewriting system
 """
 function isirreducible(w::AbstractWord, rws::RewritingSystem)
     for (lhs, _) in rules(rws)
-        occursin(lhs, w) && lhs != w && return false
+        occursin(lhs, w) && return false
     end
     return true
 end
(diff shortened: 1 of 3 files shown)


kalmarek:
let me know when ready (to me it looks ready enough:)

mikolajpabiszczak:
I want to check one more thing about it, as to which I have some doubt and have to think about it. Will do it till the end od the day.

mikolajpabiszczak:
OK, it works correctly now. The problem was with `&& lhs != w &&` in the loop: it should not be there. We guarantee that no rule is checked with itself already in the looping of `getirrsubsys` as we take only proper subwords of the rules. This additional condition was clearly wrong.
```
function isirreducible(w::AbstractWord, rws::RewritingSystem)
    for (lhs, _) in rules(rws)
        issubword(lhs, w) && lhs != w && return false
    end
    return true
end
```

Feel free to merge if you are OK with the code.

mikolajpabiszczak:
Although there is one more thing I didn't know how to do as it generated problems with use of `views` of word in `getirrsubsys`. In general one would like to have
```
function isirreducible(w::W, rws::RewritingSystem{W}) where {W}
```
(instead of what we have now) in order to guarantee that we rewrite words with respect to a `RewritingSystem` that is meant for that type of words. But doing so raises problems as `W` is different than `SubWord` which are used by `view`.

kalmarek:
thanks!

PR #23, opened 2020-07-21
Enh/bufferwords

kalmarek:


kalmarek:
benchmark:
```julia
using KnuthBendix
using Random
using Test, BenchmarkTools

    KnuthBendix.knuthbendix1(R)

    @time KnuthBendix.knuthbendix1(R)
    @time rws = KnuthBendix.knuthbendix2(R)

    Random.seed!(seed)
    n = length(rws.order.A.alphabet)
    w = Word(rand(1:n, 2000))

end;
```
and the results:
```julia
# master:
  0.407679 seconds (19.87 M allocations: 620.984 MiB, 16.91% gc time)
  0.334149 seconds (18.77 M allocations: 586.570 MiB, 21.19% gc time)
  8.730 ms (543115 allocations: 17.33 MiB)

# this brach:
  0.066934 seconds (213.23 k allocations: 8.600 MiB, 6.46% gc time)
  0.051491 seconds (114.66 k allocations: 4.732 MiB)
  1.284 ms (6 allocations: 8.30 KiB)

```
~6×faster

src/bufferwords.jl
```diff
@@ -0,0 +1,180 @@
...
+Base.@propagate_inbounds function Base.setindex!(bw::BufferWord, val, n::Integer)
+    @boundscheck checkbounds(bw, n)
+    return @inbounds bw.storage[bw.lidx+n-1] = val
+end
+Base.length(bw::BufferWord) = length(bw.lidx:bw.ridx)
+
+function Base.push!(bw::BufferWord, k::Integer)
+    if internal_length(bw) == bw.ridx
+        _growatend!(bw, max(length(bw), 16))
+    end
```
bprzybylski, line 70: In case of dynamic arrays, when the current array is full, we double the capacity of such an array. I see that you decided to extend the array by at least 16 cells. Is the reason that the words are usually short?

kalmarek: It's just a constant to balance number of additional allocations vs. memory over-conscription. We'd need to tune it or maybe even see if it makes any difference.

src/bufferwords.jl
```diff
@@ -0,0 +1,180 @@
...
+        bw = new{T}(storage, freeatbeg+1, freeatbeg + length(v))
+
+        # placing content in the middle
+        # storage[freeatbeg+1:end-freeatend] .= v
+        @inbounds for i in 1:length(v)
+            bw[i] = v[i]
+        end
+
+        return bw
+    end
```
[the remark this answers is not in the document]
kalmarek: happy to see your take on this ;) Let's discuss on Tuesday!

src/kbs2.jl
```diff
@@ -53,47 +53,42 @@ function forceconfluence!(rs::RewritingSystem, stack, i::Integer, j::Integer, o:
...
 """
-function knuthbendix2!(rs::RewritingSystem, o::Ordering = ordering(rs); maxrules::Integer = 100)
-    stack = empty(rs)
-    tmprs = empty(rs)
-    for (lhs, rhs) in rules(rs)
-        push!(stack, lhs => rhs)
-        deriverule!(tmprs, stack)
-    end
+function knuthbendix2!(rws::RewritingSystem, o::Ordering = ordering(rws); maxrules::Integer = 100)
+    stack = copy(rules(rws)[active(rws)])
```
mikolajpabiszczak, line 64: I believe this is wrong in general. In the former the use of `deriverule!(tmprs, stack)` was done after pushing each rule from original RewritingSystem to the stack. I am almost sure that this makes a huge difference, but will have to think a bit for the particular test showing that this is wrong.

src/words.jl
```diff
@@ -95,59 +14,56 @@ If type is not specified in the constructor it will default to `Int16`.
...
-    function Word{T}(v::AbstractVector{<:Integer}) where {T}
-        @assert all(x -> x > 0, v) "All entries of a Word must be positive integers"
+    function Word{T}(v::AbstractVector{<:Integer}, check=true) where {T}
+        check && @assert all(x -> x > 0, v) "All entries of a Word must be positive integers"
         return new{T}(v)
     end
     Word{T}() where T = new{T}(T[])
 end
 
 # setting the default type to Int16
```
mikolajpabiszczak, line 24: Typo - should be `UInt16`

src/kbs2.jl
```diff
@@ -13,17 +13,18 @@ function deriverule!(rs::RewritingSystem, stack, o::Ordering = ordering(rs))
...
+                isactive(rs, i) || continue
+                if occursin(a, lhs)
                     setinactive!(rs, i)
                     push!(stack, lhs => rhs)
                 else occursin(a, rhs)
                     rules(rs)[i] = (lhs => rewrite_from_left(rhs, rs))
                 end
             end
+
+            push!(rs, a => b)
```
mikolajpabiszczak, line 27: Again I think this is wrong. In the former one we first did `push!(rs, a=> b)` and then inside the for-loop we did rewriting wrt. to the RewritingSystem with that additional rule. A priori this yield different results.

src/kbs2.jl
kalmarek: Indeed I changed this, since I believe the former was incorrect, or rather non-optimal; if you investigate the old behavior it turns out that `deriverule!` is only called with `length(stack) == 1`!

src/kbs2.jl
kalmarek: it now seems to me that `else` statement is incorrect; what we should do is to first rewrite `rhs → rhs_new` with rule `a => b` (we know that this is the only rewriting rule that applies at the moment), then we need to add `a=>b` to `rs` and finally call `rewrite_from_left(rhs, rs)`.

mikolajpabiszczak: I think it will still be incorrect: noticing that in the beginning `push!(rs, a => b)` was outside the loop I believe your version may happen to push the the rule `a => b` multiple times to the RewritingSystem (because it may happen that `else` will be triggered multiple times in the for-loop).

src/kbs2.jl
mikolajpabiszczak: I looked again into it. You are right - everything works fine in that case.

src/kbs2.jl
kalmarek: I thought rather of something like this:
```julia
function rewrite_from_left!(
    v::AbstractWord,
    w::AbstractWord,
    rule::Pair{<:AbstractWord, <:AbstractWord},
)
    lhs, rhs = rule
    while !isone(w)
        push!(v, popfirst!(w))
        if issuffix(lhs, v)
            prepend!(w, rhs)
            resize!(v, length(v) - length(lhs))
        end
    end
    return v
end

function rewrite_from_left(u::W, rule::Pair{<:AbstractWord, <:AbstractWord}) where {W<:AbstractWord}
    T = eltype(u)
    v = BufferWord{T}(0, length(u))
    w = BufferWord{T}(u, 0, 0)
    v = rewrite_from_left!(v, w, rule)
    return W(v)
end
```

then the `else` block (btw. there should be `elseif`, instead of `else`!)
```julia
                elseif occursin(a, rhs)
                    rule = a => b
                    new_rhs = rewrite_from_left(rhs, rule)
                    last(rules(rs)) != rule && push!(rs, rule)
                    rules(rs)[i] = (lhs => rewrite_from_left(new_rhs, rs))
                end
```

mikolajpabiszczak: Ahhh. Then its OK :D

mikolajpabiszczak (approved):

PR #27, opened 2020-08-25
Index automata

mikolajpabiszczak:
The implementation of automata is focused more on the Index Automata than on the general automata. I did not use linked lists. I did though got inspired by them. The code extensively uses circular references. That way it is easy to walk through automata (both forward and backward). The drawback is that now modification of automaton is somewhat difficult (one has to modify some number of fields).

### DIFF
--- a/src/automata.jl
+++ b/src/automata.jl
@@ -1,19 +1,17 @@
-
-
 abstract type AbstractState{T} end
 
 mutable struct State{T} <: AbstractState{T}
-    name::Union{Word{T}, SubWord{T}}
+    name::AbstractWord{T}
     term::Bool
-    rrule::Union{Word{T}, SubWord{T}, Nothing}
+    rrule::Union{AbstractWord{T}, Nothing}
     ined::Vector{Union{State{T}, Nothing}}
     outed::Vector{Union{State{T}, Nothing}} # LinkedList?
 end
 
 
-State(name::Union{Word{T}, SubWord{T}}, absize::Int) where {T} = State(name, false, nothing, Vector{Union{State{T}, Nothing}}(nothing, absize), Vector{Union{State{T}, Nothing}}(nothing, absize))
-State(name::Union{Word{T}, SubWord{T}}, rrule::Union{Word{T}, SubWord{T}, Nothing}) where T = State(name, true, rrule, State{T}[], State{T}[])
-State(name::Union{Word{T}, SubWord{T}}) where T = State(name, false, nothing, State{T}[], State{T}[])
+State(name::AbstractWord{T}, absize::Int) where {T} = State(name, false, nothing, Vector{Union{State{T}, Nothing}}(nothing, absize), Vector{Union{State{T}, Nothing}}(nothing, absize))
+State(name::AbstractWord{T}, rrule::Union{AbstractWord{T}, Nothing}) where T = State(name, true, rrule, State{T}[], State{T}[])
+State(name::AbstractWord{T}) where T = State(name, false, nothing, State{T}[], State{T}[])
 
 inedges(s::State) = s.ined
 outedges(s::State) = s.outed
@@ -22,29 +20,36 @@ rightrule(s::State) = s.rrule
 name(s::State) = s.name
 Base.length(s::State) = length(name(s))
 
+function declarerightrule!(s::State{T}, rrule::AbstractWord{T}) where {T}
+    rightrule(s) = rrule
+    isterminal(s) = true
+end
+
 
 abstract type AbstractAutomaton end
 
 mutable struct Automaton{T} <: AbstractAutomaton
     states::Vector{State{T}} # Linked List?  # first element is initial state
+    nameidx::Vector{<:AbstractWord{T}} # To facilitate the fast search over names
     abt::Alphabet
 end
 
 # setting the default type to UInt16
-Automaton(states::Vector{Union{State{T}, AbstractState{T}}}, abt::Alphabet) where {T<:Integer} = Automaton{UInt16}(states, abt)
+Automaton(states::Vector{Union{State{T}, AbstractState{T}}}, idx::Vector{AbstractWord{T}}, abt::Alphabet) where {T<:Integer} = Automaton{UInt16}(states, idx, abt)
 
-Automaton(abt::Alphabet) = Automaton([State(Word(Int[]), length(abt))], abt)
+Automaton(abt::Alphabet) = Automaton([State(Word(Int[]), length(abt))], [Word(Int[])], abt)
 
 states(a::Automaton) = a.states
+names(a::Automaton) = a.nameidx
 initialstate(a::Automaton) = states(a)[1]
 
 function Base.push!(a::Automaton{T}, s::State{T}) where {T}
     @assert (length(s.ined) == length(a)) && (length(s.outed) == length(a)) "Tables of in and out edges must have the same length as alphabet"
-    push!(a.states, s)
+    push!(a.states, s); push!(names(a), name(s))
 end
 
-function Base.push!(a::Automaton{T}, name::Union{Word{T}, SubWord{T}}) where {T}
-    push!(states(a), State(name, length(a.abt)))
+function Base.push!(a::Automaton{T}, name::AbstractWord{T}) where {T}
+    push!(states(a), State(name, length(a.abt))); push!(names(a), name)
 end
 
 function addedge!(a::Automaton, letter::Int, from::Int, to::Int)
@@ -54,7 +59,25 @@ function addedge!(a::Automaton, letter::Int, from::Int, to::Int)
     outedges(states(a)[from])[letter] = states(a)[to]
     inedges(states(a)[to])[letter] = states(a)[from]
 end
+
+function addedge!(a::Automaton{T}, letter::Int, from::State{T}, to::State{T}) where {T}
+    @assert from in states(a) "State from which the edge is added must be inside automaton"
+    @assert to in states(a) "State to which the edge is added must be inside automaton"
+    outedges(from)[letter] = to
+    inedges(to)[letter] = from
+end
 # Should we also check if the edge corresponding to a given letter does not exist?
+
+# function addpath!(a::Automaton{T}, w::AbstractWord{T}) where {T}
+#     σ = states(a)[1]
+#     for (i, letter) in enumerate(w)
+#         if σ[letter] === nothing  # this is wrong condition
+#             push!(a, @view(w[1:i]))
+#             addedge!(a, letter, σ, states(a)[end])
+#         else σ = σ[letter]
+#         end
+#     end
+# end
 
 function removeedge!(a::Automaton, letter::Int, from::Int, to::Int)
     @assert 0 < letter <= length(a.abt) "Edge must be a valid pointer to the letter"
@@ -66,11 +89,11 @@ end
 
 
 function Base.deleteat!(a::Automaton, idx::Int)
-    for (i, state) in enumerate(outedges(states(a)[idx]))
-        inedges(state)[i] = undef
+    for (i, σ) in enumerate(outedges(states(a)[idx]))
+        inedges(σ)[i] = undef
     end
-    for (i, state) in enumerate(inedges(states(a)[idx]))
-        outedges(state)[i] = undef
+    for (i, σ) in enumerate(inedges(states(a)[idx]))
+        outedges(σ)[i] = undef
     end
     deleteat!(states, idx)
 end
@@ -78,12 +101,12 @@ end
 
 
 function walk(a::Automaton{T}, w::Union{Word{T}, SubWord{T}}, first::Int) where {T}
-    current = states(a)[first]
+    σ = states(a)[first]
     for i in w
-        next = outedges(current)[i]
-        next === nothing ? error("No path corresponding to a given word exists") :  current = next
+        next = outedges(σ)[i]
+        next === nothing ? error("No path corresponding to a given word exists") :  σ = next
     end
-    return current
+    return σ
 end
 
 walk(a::Automaton{T}, w::Union{Word{T}, SubWord{T}}) where{T} = walk(a, w, 1)
@@ -121,6 +144,48 @@ Base.length(abt::Alphabet) = length(abt.alphabet)
 
 # Move to rewriting
 
+function index(rws::RewritingSystem, abt::Alphabet)
+    Σᵢ = Int[0]
+    a = Automaton(abt)
+    # Determining simple paths
+    for (lhs, rhs) in rules(rws)
+        σ = a[1]
+        for (i, letter) in enumerate(lhs)
+            if !isnothing(outedges(σ)[letter])
+                σ = outedges(σ)[letter]
+            else
+                push!(a, lhs[1:i])
+                push!(Σᵢ, i)
+                addedge!(a, σ, a[end])
+            end
+        end
+        terminal = a[end]
+        declarerightrule!(terminal, rhs)
+        # Add loops for terminal state (not needed for rewriting, so commenting out)
+        # for letter in outedges(terminal)
+        #     addedge!(a, letter, terminal, terminal)
+        # end
+    end
+    # Determining cross paths
+    for letter in outedges(a[1])
+        isnothing(letter) && addedge!(a, letter, 1, 1)
+    end
+    i = 1
+    indcs = findall(isequal(i), Σᵢ)
+    while !isempty(indcs)
+        for idx in indcs
+            σ = a[idx]
+            τ = walk(a, name(σ)[2:end])
+            for letter in outedges(σ)
+                isnothing(letter) && addedge!(a, letter, σ, τ)
+            end
+        end
+        i += 1
+        indcs = findall(isequal(i), Σᵢ)
+    end
+end
+
+
 function index_rewrite(u::AbstractWord{T}, a::Automaton{T}, rws::RewritingSystem) where {T}
     v = one(u)
     w = copy(u)
@@ -140,5 +205,6 @@ function index_rewrite(u::AbstractWord{T}, a::Automaton{T}, rws::RewritingSystem
             w = prepend!(w, rightrule(state))
         end
     end
+    return v
 end
 
--- a/src/automata.jl
+++ b/src/automata.jl
@@ -2,10 +2,10 @@
 # States
 
 """
-    AbstractState{T, N}
+    AbstractState{T, N, W}
 Abstract type representing state of the (index) automaton.
 
-The subtypes of `AbstractState{T, N}` should implement the following methods which
+The subtypes of `AbstractState{T, N, W}` should implement the following methods which
 constitute `AbstractState` interface:
  * `name`: gives external name of the state
  * `isterminal`: returns whether the state is terminal or not (i.e. whether the
@@ -22,10 +22,10 @@ constitute `AbstractState` interface:
  * `Base.length`: gives the length of the signature of the shortest path to the
     state, which starts at initial state (length of "simple path")
 """
-abstract type AbstractState{T, N} end
+abstract type AbstractState{T, N, W} end
 
 """
-    State{T, N} <: AbstractState{T, N}
+    State{T, N, W<:AbstractWord{T}} <: AbstractState{T, N, W}
 State as a name (corresponding to the signature of the simple path ending at this
 state), indication whether it is terminal (in which case there is a right-hand part
 of the rewriting rule, vector of states (indexed by the position of the letter in
@@ -33,26 +33,26 @@ the alphabet) from which there is an edge (labelled by letter indexed) to this s
 and vector of states (indexed by position of the letter in the alphabet) to which
 there is an edge (labelled by letter indexed) from this state.
 """
-mutable struct State{T, N} <: AbstractState{T, N}
-    name::Word{T}
+mutable struct State{T, N, W<:AbstractWord{T}} <: AbstractState{T, N, W}
+    name::W
     terminal::Bool
-    rrule::Word{T}
-    ined::Vector{State{T, N}}
-    outed::NTuple{N, State{T, N}}
+    rrule::W
+    ined::Vector{State{T, N, W}}
+    outed::NTuple{N, State{T, N, W}}
     representsnoedge::Bool
 
-    function State{T, N}() where {T, N}
+    function State{T, N, W}() where {T, N, W<:AbstractWord{T}}
         x = new()
         x.representsnoedge = true
         return x
     end
 
-    function State(name::AbstractWord{T}, noedge::State{T, N}) where {T, N}
-        s = State{T, N}()
+    function State(name::W, noedge::State{T, N, W}) where {T, N, W<:AbstractWord{T}}
+        s = State{T, N, W}()
         s.name = name
         s.terminal = false
-        s.rrule = Word{T}()
-        s.ined = State{T, N}[]
+        s.rrule = W()
+        s.ined = State{T, N, W}[]
         s.outed = ntuple(i -> noedge, N)
         s.representsnoedge = false
         return s
@@ -68,11 +68,11 @@ isnoedge(s::State) = s.representsnoedge
 Base.length(s::State) = length(name(s))
 
 """
-    declarerightrule!(s::State, w::AbstractWord)
+    declarerightrule!(s::State{T, N, W}, w::W) where {T, N, W}
 Decalres given word as a right-hand side rule of a given state (and makes this)
 state terminal.
 """
-function declarerightrule!(s::State, w::AbstractWord)
+function declarerightrule!(s::State{T, N, W}, w::W) where {T, N, W}
     s.terminal = true
     s.rrule = w
 end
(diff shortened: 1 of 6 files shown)


mikolajpabiszczak:
Just to compare:

```
rsc = KnuthBendix.RewritingSystem([a=>ε, b=>ε, c=>ε, d=>ε, ba=>ab, be=>eb, pa=>ap, pe=>ep], lenlexord)
ia = KnuthBendix.makeindexautomaton(rsc, A)

seed=1
Random.seed!(seed)
n = length(A)
w = Word(rand(1:n, 2000))

@btime KnuthBendix.rewrite_from_left(w, rsc)
@btime KnuthBendix.index_rewrite(w, ia)
```

yields

```
# rewrite_from_left(w, rsc)
7.367 ms (565998 allocations: 20.83 MiB)

# KnuthBendix.index_rewrite(w, ia)
4.775 ms (41078 allocations: 2.08 MiB)
```

src/automata.jl
```diff
@@ -0,0 +1,291 @@
...
+    while !isone(w)
+        x = popfirst!(w)
+        k = length(v) + 1
+        state = outedges(states[k])[x]
+
+        if !isterminal(state)
+            @inbounds states[k+1] = state
+            push!(v, x)
+        else
+            v = v[1:end-length(state)+1]
```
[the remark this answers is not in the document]
kalmarek: Actually let's wait for `BufferWord` to arrive

kalmarek:
also a more meaningful benchmark:
This branch:
```julia
julia> using KnuthBendix
[ Info: Precompiling KnuthBendix [c2604015-7b3d-4a30-8a26-9074551ec60a]

julia> using Random

julia> using Test, BenchmarkTools

julia> include("test/rws_examples.jl")
Test Summary: | Pass  Total
KBS1 examples |    8      8
Test Summary: | Pass  Total
KBS2 examples |    8      8
Test.DefaultTestSet("KBS2 examples", Any[], 8, false)

julia> let R = RWS_Example_237_abaB(4), seed=1
           KnuthBendix.knuthbendix1(R)

           @time KnuthBendix.knuthbendix1(R)
           @time rws = KnuthBendix.knuthbendix2(R)

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

           ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
           @btime KnuthBendix.index_rewrite($w, $ia)
       end;
┌ Warning: Maximum number of rules (100) in the RewritingSystem reached.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs1.jl:94
┌ Warning: Maximum number of rules (100) in the RewritingSystem reached.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs1.jl:94
  0.305695 seconds (9.18 M allocations: 421.955 MiB, 9.50% gc time)
  0.220560 seconds (8.74 M allocations: 401.247 MiB, 12.55% gc time)
  6.631 ms (222697 allocations: 10.24 MiB)
  1.331 ms (8670 allocations: 317.84 KiB)
```

vs. BufferWords:
```julia
julia> using KnuthBendix
[ Info: Precompiling KnuthBendix [c2604015-7b3d-4a30-8a26-9074551ec60a]

julia> using Random

julia> using Test, BenchmarkTools

julia> include("test/rws_examples.jl")
┌ Warning: Maximum number of rules (100) in the RewritingSystem reached.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs1.jl:98
Test Summary: | Pass  Broken  Total
KBS1 examples |    7       2      9
Test Summary: | Pass  Total
KBS2 examples |   10     10
Test.DefaultTestSet("KBS2 examples", Any[], 10, false)

julia> let R = RWS_Example_237_abaB(4), seed=1
           KnuthBendix.knuthbendix1(R)

           @time KnuthBendix.knuthbendix1(R)
           @time rws = KnuthBendix.knuthbendix2(R)

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

       end;
┌ Warning: Maximum number of rules (100) in the RewritingSystem reached.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs1.jl:98
┌ Warning: Maximum number of rules (100) in the RewritingSystem reached.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs1.jl:98
  0.064033 seconds (73.50 k allocations: 5.075 MiB)
  0.054629 seconds (22.11 k allocations: 1.755 MiB, 17.71% gc time)
  1.336 ms (5 allocations: 8.30 KiB)
```
This actually looks good, if we could achieve <0.5ms combining both I'd be pretty happy ;)

mikolajpabiszczak:
> 1. Do i understand correctly, that each word (in node/state) is stored `(indegree+outdegree+1)`-times? that's hugely ineffective

I am not sure what you mean by "word" - do you mean a state? States (corresponding to a word). I believe you are asking about the craziness caused by the circular references to states in `inedges` and `outedges` - I believe (or at least I hope it is true) Julia does not store the states in those vectors again, it just points to the place in memory where they are stored. This is inspired by the way linked lists are implemented in DataStructures.jl.

One can get rid of `inedges` but this will make removing nodes later more difficult - BUT (and here is something important): we will have to decide what should function as a `RewritingSystem` because right know KBS1/2 creates and operates on `RewritingSystem` but it seems we would like to use `IndexAutomaton` instead. Then we would like to have some sensible way of removing states and edges from automata, hence we should leave  `inedges`.

> 2. is is really necessary to make fields in `State`s  `Union{..., Nothing}`? those introduce noticeable slowdown. Does it really happen that in/out-edges contain states and `nothing` at the same time? for nothingness of `rrule` we should just add `terminal::Bool` field and store `one(name)`.

Yes, it happens that we store both `Nothing` and a `State`.  Actually it happens most of the time. Let me first sketch what is the idea behind `inedges` and `outedges`. I wanted them to have a constant size equal to the size of the alphabet - that way the `in/outedges` are bascially indexed by letters (say I want to get from state "something" using the edge with label "second letter of the alphabet" - I have `something.outedges[2]`. If I were to store only existing edges then I would first need to search through the list of edges to find the proper edge and then go to the state it points.

I believe your question is based on the view of the `IndexAutomaton` as constructed by `makeindexautomaton`. Note that during the course of constructing this automaton `in/outedges` of states inside automaton have plenty of `Nothing`s the job of `makeindexautomaton` is to get rid of them. If we want to get rid of `Nothing` then one can think about creating kind of a `NothingState <: AbstractState` or something like that, but I did not have an idea how to do that.

For nothingness of `rrule` you are right.

src/automata.jl
```diff
@@ -0,0 +1,291 @@
...
+abstract type AbstractAutomaton end
+
+"""
+    Automaton{T} <: AbstractAutomaton
+Automaton as a vector of states together with alphabet.
+"""
+mutable struct Automaton{T} <: AbstractAutomaton
+    states::Vector{State{T}}
+    abt::Alphabet
+end
```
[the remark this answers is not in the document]
mikolajpabiszczak: You are right. It needs not to.

src/automata.jl
```diff
@@ -0,0 +1,291 @@
...
+initialstate(a::Automaton) = states(a)[1]
+
+function Base.push!(a::Automaton{T}, s::State{T}) where {T}
+    @assert (length(s.ined) == length(a)) && (length(s.outed) == length(a)) "Tables of in and out edges must have the same length as alphabet"
+    push!(a.states, s)
+end
+
+function Base.push!(a::Automaton{T}, name::AbstractWord{T}) where {T}
+    push!(states(a), State(name, length(a.abt)))
+end
```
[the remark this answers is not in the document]
mikolajpabiszczak: Done.

kalmarek:
> > 1. Do i understand correctly, that each word (in node/state) is stored `(indegree+outdegree+1)`-times? that's hugely ineffective
> 
> I am not sure what you mean by "word" - do you mean a state? States (corresponding to a word). I believe you are asking about the craziness caused by the circular references to states in `inedges` and `outedges` - I believe (or at least I hope it is true) Julia does not store the states in those vectors again, it just points to the place in memory where they are stored. This is inspired by the way linked lists are implemented in DataStructures.jl.

If that's indeed true, then I'm happy with this solution; Please check this is indeed true;
 
> One can get rid of `inedges` but this will make removing nodes later more difficult - BUT (and here is something important): we will have to decide what should function as a `RewritingSystem` because right know KBS1/2 creates and operates on `RewritingSystem` but it seems we would like to use `IndexAutomaton` instead. Then we would like to have some sensible way of removing states and edges from automata, hence we should leave `inedges`.

If only pointers are stored, there is no need to remove `inedges`; In the long run we (probably?) want to incorporate `IndexAutomaton` into `RewritingSystem`.

> > 1. is is really necessary to make fields in `State`s  `Union{..., Nothing}`? those introduce noticeable slowdown. Does it really happen that in/out-edges contain states and `nothing` at the same time? for nothingness of `rrule` we should just add `terminal::Bool` field and store `one(name)`.
> 
> Yes, it happens that we store both `Nothing` and a `State`. Actually it happens most of the time. Let me first sketch what is the idea behind `inedges` and `outedges`. I wanted them to have a constant size equal to the size of the alphabet - that way the `in/outedges` are bascially indexed by letters (say I want to get from state "something" using the edge with label "second letter of the alphabet" - I have `something.outedges[2]`. If I were to store only existing edges then I would first need to search through the list of edges to find the proper edge and then go to the state it points.

Thanks for the explanation, I thought these were "sparse lists"; but I like the design ;)

> EDIT: BTW: we do not need to store alphabet inside the automaton. It was kind of a thing I added to easily get constant size of `in/outedges` without using some "magic numbers".

If the size is known and fixed at compile time, then we should use `NTuple{N, W}` instead of `Vectors` and add `N` to the type parameters of `IndexAutomaton`; Since `nothing` represents "no_edge", then again it's better to add a field representing this function in `State` and produce a possibly minimal in size (non-initialized) `State{W, N}`. Type uniform data structures are much easier for llvm to optimize. (Though small unions are getting better and better).

it's looking great!

mikolajpabiszczak:
Ad `in/outedges` first some general check:

```julia
julia> a=[1,2,3]; b=[2,3,4]; c=[3,4,5];

julia> d=[a,b,c]
3-element Array{Array{Int64,1},1}:
 [1, 2, 3]
 [2, 3, 4]
 [3, 4, 5]

julia> pointer(a) == pointer(d[1])
true

julia> pointer(b) == pointer(d[2])
true

julia> pointer(c) == pointer(d[3])
true

julia> a[3] = 9
9

julia> d
3-element Array{Array{Int64,1},1}:
 [1, 2, 9]
 [2, 3, 4]
 [3, 4, 5]
```

Now in our case everything behaves in the similar way:
```julia
using KnuthBendix
abt = KnuthBendix.Alphabet(['a', 'b', 'c', 'd'])
a = KnuthBendix.Automaton(abt)
push!(a, KnuthBendix.Word([1]))
KnuthBendix.addedge!(a, 1, 1, 2)

julia> KnuthBendix.inedges(KnuthBendix.states(a)[2])[1] === KnuthBendix.states(a)[1]
true
```

I did some googling - thats way I believed Julia does what it does:
https://stackoverflow.com/questions/45704082/julia-array-assignment-behaviour/45705342
https://stackoverflow.com/questions/33002572/creating-copies-in-julia-with-operator
https://stackoverflow.com/questions/38601141/what-is-the-difference-between-and-comparison-operators-in-julia

BTW: If Julia did copy the states into each array, then the whole updating (adding/removing edges) of the automaton would not work.

As to the last thing: `NTuple{N, W}` - this is immutable type and it won't work - we can't updated edges assignments.

kalmarek:
> As to the last thing: `NTuple{N, W}` - this is immutable type and it won't work - we can't updated edges assignments.

since these contain just pointers, we can just construct a new one?
The main advantage is that tuples are stack-allocated, whereas vectors reside on the heap

mikolajpabiszczak:
> we can just construct a new one?

We can. BTW: is there a simply way to create new tuple that differs from the old one just in one place/index?

kalmarek:
(name chosen randomly)
```julia
julia> function replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
           @boundscheck 1 ≤ idx ≤ N || throw(BoundsError("Tried to access $(typeof(k)) at index $idx"))
           return ntuple( i->i==idx ? val : k[i], N)
       end
replace (generic function with 1 method)

julia> t = ntuple(identity, 5)
(1, 2, 3, 4, 5)

julia> replace(t, 10, 2)
(1, 10, 3, 4, 5)
```
probably you could make it faster, but we'll see to it when it becomes the bottleneck:
```
julia> @benchmark replace(Ref($t)[], 10, 2)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.247 ns (0.00% GC)
  median time:      2.264 ns (0.00% GC)
  mean time:        2.406 ns (0.00% GC)
  maximum time:     41.172 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

```
note that `Ref($t)[]` is necessary to get correct timings, as
```
julia> @btime replace($t, 10, 2)
  0.028 ns (0 allocations: 0 bytes)
```
propagates constants and folds the function :)

mikolajpabiszczak:
> ```julia
> julia> function replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
>            @boundscheck 1 ≤ idx ≤ N || throw(BoundsError("Tried to access $(typeof(k)) at index $idx"))
>            return ntuple( i->i==idx ? val : k[i], N)
>        end
> ```

Ahh, of course. Beautiful.

BTW: Yesterday it occurred to me that the concept of `inedges` is not what I wanted it to be. `outedges` are OK because from every state there can be only one path labelled by given letter. But for the `inedges`: 1) a priori the number of entering edges can be larger than the length of alphabet (even in case of reduced rewriting systems); even worse - and I have no idea what I've been thinking earlier - 2) every single edge entering given state will be labelled by the same letter. "Brawo Ja" 🙄 Although I think that it can be fixed by taking `inedges` to be a vector where on the first position we always store the previous state on the "direct path" and on the other will be states from the cross paths. I believe it is everything needed to have a sensible deleting of states.

I will push the fixes to all the comments and the problem above later.

mikolajpabiszczak:
Still without bufferwords and it looks nice.

```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

           ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
           @btime KnuthBendix.index_rewrite($w, $ia)
        end;
  6.567 ms (543115 allocations: 17.33 MiB)
  509.947 μs (4594 allocations: 234.56 KiB)
```

src/automata.jl
```diff
@@ -18,38 +18,54 @@ constitute `AbstractState` interface:
...
+    rrule::Word{T}
+    ined::Vector{State{T, N}}
+    outed::NTuple{N, State{T, N}}
+    representsnoedge::Bool
+
+    function State{T, N}() where {T, N}
+        x = new()
+        x.representsnoedge = true
+        return x
+    end
```
kalmarek: please parametrize `State{T,N,W<:AbstractWord{T}} <: AbstractState{T,N}` by the type of the word.
The parameter needs to be added to the inner constructors then as well (to avoid abstract types).

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
 end
 
-# setting the default type to UInt16
-Automaton(states::Vector{Union{State{T}, AbstractState{T}}}, abt::Alphabet) where {T<:Integer} = Automaton{UInt16}(states, abt)
-
-Automaton(abt::Alphabet) = Automaton([State(Word(Int[]), length(abt))], abt)
+# Default type is set to UInt16
+function Automaton(abt::Alphabet)
+    uniquenoedge = State{UInt16, length(abt)}()
+    Automaton{UInt16, length(abt)}([State(Word(), length(abt), uniquenoedge)], abt, uniquenoedge)
```
kalmarek, line 138:
```suggestion
    Automaton{UInt16, length(abt)}([State(Word{UInt16}(), length(abt), uniquenoedge)], abt, uniquenoedge)
```

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
+Base.push!(a::Automaton{T}, name::AbstractWord{T}) where {T} = push!(states(a), State(name, length(a.abt), noedge(a)))
 
-function Base.push!(a::Automaton{T}, s::State{T}) where {T}
-    @assert (length(s.ined) == length(a)) && (length(s.outed) == length(a)) "Tables of in and out edges must have the same length as alphabet"
-    push!(a.states, s)
+"""
+    replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
+Returns a copy of `NTuple` `k` with value at index `idx` changed to `val`.
+"""
+function replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
```
kalmarek, line 152:
```suggestion
Base.@propagate_inbounds function replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
```

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
+    updateoutedges!(s::State, to::State, label::Integer)
+Updates outedges of the state `s` with state `to` connected through the edge `label`.
+"""
+function updateoutedges!(s::State, to::State, label::Integer)
+    s.outed = replace(outedges(s), to, label)
+end
+
+"""
+    addinedge!(s::State, from::State)
+Updates inedges of the state `s` with state `from` connected through the edge `label`.
```
kalmarek, line 168: what is `label`?

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
+    uniquenoedge = State{UInt16, length(abt)}()
+    Automaton{UInt16, length(abt)}([State(Word(), length(abt), uniquenoedge)], abt, uniquenoedge)
+end
 
 states(a::Automaton) = a.states
 initialstate(a::Automaton) = states(a)[1]
+noedge(a::Automaton) = a.uniquenoedge
+
+Base.push!(a::Automaton{T}, s::State{T}) where {T} = push!(a.states, s)
+Base.push!(a::Automaton{T}, name::AbstractWord{T}) where {T} = push!(states(a), State(name, length(a.abt), noedge(a)))
```
kalmarek, line 146:
```suggestion
Base.push!(a::Automaton{T, N}, name::AbstractWord{T}) where {T, N} = push!(states(a), State(name, N, noedge(a)))
```

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
 
-function Base.push!(a::Automaton{T}, s::State{T}) where {T}
-    @assert (length(s.ined) == length(a)) && (length(s.outed) == length(a)) "Tables of in and out edges must have the same length as alphabet"
-    push!(a.states, s)
+"""
+    replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
+Returns a copy of `NTuple` `k` with value at index `idx` changed to `val`.
+"""
+function replace(k::NTuple{N, T}, val, idx::Integer) where {N, T}
+    @boundscheck 1 ≤ idx ≤ N || throw(BoundsError("Tried to access $(typeof(k)) at index $idx"))
```
kalmarek, line 153: I actually didn't test what I wrote:) this is correct
```suggestion
    @boundscheck 1 ≤ idx ≤ N || throw(BoundsError(k, idx))
```

src/automata.jl
```diff
@@ -102,50 +120,76 @@ constitute `AbstractAutomaton` interface:
...
+function addinedge!(s::State, from::State)
+    push!(inedges(s), from)
+end
 
 """
     addedge!(a::Automaton, label::Integer, from::Integer, to::Integer)
-Adds the edge with a given `label` `from` one state `to` another .
+Adds the edge with a given `label` directed `from` state `to` state.
 """
 function addedge!(a::Automaton, label::Integer, from::Integer, to::Integer)
```
kalmarek, line 178: * add `Base.@propagate_inbounds` and replace `@asserts` by `@boundscheck checkbounds(a.abt, label)`, etc
* prepend `@inbounds` to all lines 182-185. we have checked above that these should be safe. In general we should make sure that all of those operations are `@inbounds` after the check.

also: `length(a.abt)` → `N`

src/automata.jl
```diff
@@ -155,18 +199,22 @@ Removes the edge with a given `label` `from` one state `to` another.
 """
 function removeedge!(a::Automaton, label::Integer, from::Integer, to::Integer)
     @assert 0 < label <= length(a.abt) "Edge must be a valid pointer to the label (letter)"
-    @assert 0 < from <= length(states(a)) "Edges can be added only between states inside automaton"
-    @assert 0 < to <= length(states(a)) "Edges can be added only between states inside automaton"
-    outedges(states(a)[from])[label] = nothing
-    inedges(states(a)[to])[label] = nothing
+    @assert 0 < from <= length(states(a)) "Edges can be removed only between states inside automaton"
```
kalmarek, line 202: as above

kalmarek (changes requested):

kalmarek:
ok, so what is the plan @mikolajpabiszczak ? should we merge this (when ready), and then integrate with RewritingSystems in a separate PR? or do you want to work on this further here?

src/automata.jl
mikolajpabiszczak: My bad. In this case we store no label. Corrected.

src/automata.jl
mikolajpabiszczak: So I `btime`'ed it and what you propose is slower:

```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

                  Random.seed!(seed)
                  n = length(rws.order.A.alphabet)
                  w = Word(rand(1:n, 2000))

                  ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
                  @btime KnuthBendix.index_rewrite($w, $ia)
               end;
  6.926 ms (543115 allocations: 17.33 MiB)
  1.212 ms (6050 allocations: 280.06 KiB)
```

vs the one I did:

```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

           ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
           @btime KnuthBendix.index_rewrite($w, $ia)
       end;
  6.933 ms (543115 allocations: 17.33 MiB)
  501.837 μs (4594 allocations: 234.56 KiB)
```

Although I am not sure if I implemented your suggestion correctly:
```julia
mutable struct State{T, N, W<:AbstractWord{T}} <: AbstractState{T, N, W}
    name::W
    terminal::Bool
    rrule::W
    ined::Vector{State{T, N, W}}
    outed::NTuple{N, State{T, N, W}}
    representsnoedge::Bool

    function State{T, N, W}() where {T, N, W}
        x = new()
        x.representsnoedge = true
        return x
    end

    function State(name::W, absize::Integer, noedge::State{T, N, W}) where {T, N, W}
        @assert N == absize "Length of the alphabet does not match!"
        s = State{T, N, W}()
        s.name = name
        s.terminal = false
        s.rrule = W()
        s.ined = State{T, absize}[]
        s.outed = ntuple(i -> noedge, absize)
        s.representsnoedge = false
        return s
    end
end
```

kalmarek: most probably there is something abstractly typed now in the code:
```
1.212 ms (6050 allocations: 280.06 KiB)
```
vs
```
501.837 μs (4594 allocations: 234.56 KiB)
```
adding parameter doesn't produce new allocations

kalmarek:
```
s.ined = State{T, absize}[]
```
is abstractly typed, you need `State{T, N, W}[]`

mikolajpabiszczak: It required slightly more work but here it is:

```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

           ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
           @btime KnuthBendix.index_rewrite($w, $ia)
       end;
  6.877 ms (543115 allocations: 17.33 MiB)
  218.877 μs (4594 allocations: 234.56 KiB)
```

mikolajpabiszczak:
I think this is worth putting in the main thread (not only in the change suggestion one):

```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           Random.seed!(seed)
           n = length(rws.order.A.alphabet)
           w = Word(rand(1:n, 2000))

           ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.ordering(rws).A)
           @btime KnuthBendix.index_rewrite($w, $ia)
       end;
  6.877 ms (543115 allocations: 17.33 MiB)
  218.877 μs (4594 allocations: 234.56 KiB)
```

And this is still without bufferwords! Please look if you have any other suggestions. If not then I believe it can be merged and then we may incorporate optimisations from the bufferwords.

kalmarek (approved):

kalmarek:
lgtm!

PR #30, opened 2020-09-16
Enh/rewrite interface

kalmarek:
* add uniform `string_repr(w::AbstractWord, context)` where `context` could be either `Alphabet` or `Ordering` 
* cleaned-up `rewrite_from_left`
* simplified lots of signatures from `automata.jl`
* integrated `Automaton` and its rewriting into it
* some performance tunning, notably: added `_past_states` field to `Automaton` to be re-used in `rewrite_from_left`

```julia
using Test, BenchmarkTools, Random
using KnuthBendix
```
before:
```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           n = length(KnuthBendix.alphabet(KnuthBendix.ordering(rws)))
           Random.seed!(seed)
           w = Word(rand(1:n, 2000))

           @time ia = KnuthBendix.makeindexautomaton(rws, KnuthBendix.alphabet(KnuthBendix.ordering(rws)))
           @btime KnuthBendix.index_rewrite($w, $ia)
       end;
  1.453 ms (5 allocations: 8.30 KiB)
  0.000742 seconds (4.76 k allocations: 218.672 KiB)
  242.169 μs (4430 allocations: 229.52 KiB)
```

after
```julia
julia> let R = RWS_Example_237_abaB(4), seed=1

           n = length(KnuthBendix.alphabet(KnuthBendix.ordering(rws)))
           Random.seed!(seed)
           w = Word(rand(1:n, 2000))

           @time ia = KnuthBendix.makeindexautomaton(rws)
           @btime KnuthBendix.rewrite_from_left($w, $ia)
       end;
  1.170 ms (5 allocations: 8.30 KiB)
  0.000483 seconds (4.76 k allocations: 218.750 KiB)
  98.662 μs (5 allocations: 8.30 KiB)
```

### DIFF
--- a/src/automata.jl
+++ b/src/automata.jl
@@ -128,12 +128,13 @@ struct Automaton{T, N, W<:AbstractWord{T}} <: AbstractAutomaton{T, N, W}
     states::Vector{State{T, N, W}}
     abt::Alphabet
     uniquenoedge::State{T, N, W}
+    _past_states::Vector{State{T, N, W}}
 end
 
 # Default type is set to UInt16
 function Automaton(abt::Alphabet)
     uniquenoedge = State{UInt16, length(abt), Word{UInt16}}()
-    Automaton{UInt16, length(abt), Word{UInt16}}([State(Word{UInt16}(), uniquenoedge)], abt, uniquenoedge)
+    Automaton{UInt16, length(abt), Word{UInt16}}([State(Word{UInt16}(), uniquenoedge)], abt, uniquenoedge, State{UInt16, length(abt), Word{UInt16}}[])
 end
 
 alphabet(a::Automaton) = a.abt
@@ -304,7 +305,8 @@ Rewrites word `w` from left using index automaton `a` and appends the result
 to `v`. For standard rewriting `v` should be empty.
 """
 function rewrite_from_left!(v::AbstractWord, w::AbstractWord, a::Automaton)
-    past_states = similar(states(a), length(w))
+    past_states = a._past_states
+    resize!(past_states, length(w))
     state = initialstate(a)
     past_states[1] = state
     initial_length = length(v)
--- a/src/automata.jl
+++ b/src/automata.jl
@@ -240,19 +240,17 @@ end
 walk(a::Automaton{T, N, W}, w::W) where {T, N, W} = walk(a, w, 1)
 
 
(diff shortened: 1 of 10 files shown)


src/automata.jl
```diff
@@ -61,18 +61,18 @@ end
 
 name(s::State) = s.name
 isterminal(s::State) = s.terminal
-rightrule(s::State) = isterminal(s) ? s.rrule : nothing
+rightrule(s::State) = s.rrule
```
mikolajpabiszczak, line 64: Frankly, I prefer the original. There is a reason I have hidden empty word behind `nothing` when the state is terminal - the reason is that the empty word may be the right rule itself (e.g. in case aa^{-1})). Note that the job of `rightrule()`is just to get the real existing rule and the only reason we would like to access the field `s.rrule` (and not get nothing) is while declaring right rule. But this is done using `declarerightrule()` (and should always be performed using that method!)

src/rewriting.jl
```diff
@@ -1,3 +1,43 @@
+"""
+    rewrite_from_left(u::AbstractWord, rewriting)
+Rewrites a word from left using `rewriting` object. The object must implement
+`rewrite_from_left!(v::AbstractWord, w::AbstractWord, rewriting)` to succesfully rewrite `u`.
+"""
+function rewrite_from_left(u::W, rewriting) where {W<:AbstractWord}
```
mikolajpabiszczak, line 6: Wouldn't it be sensible to declare additional type `RewritingObject` which would be a supertype of `AbstractRewritingSystem` and `AbstractAutomaton`?

src/automata.jl
kalmarek: In the original form this method allocates (boxed value) and as such can not be used in hot loop;
revert this change and compare timings

src/rewriting.jl
kalmarek: sure, why not; but I see little (no?) added value with this abstraction, except making the signature clearer;
I hope the docstring is clear enough?

On the other hand if someones `rewriting_object` is already a subtype of a distinct type hierarchy we'd effectively force to define a shim around it to be used here. That's far fetched, so that I wouldn't count this as an argument here though.

mikolajpabiszczak (approved):

PR #31, opened 2020-09-22
Basic wreath-product ordering implemented

mikolajpabiszczak:
Ad #21 
I know that I named it in a weird way, but this is the name that occurs in Sims. Please say if you prefer `WreathOrder`.

### DIFF
--- a/src/orderings.jl
+++ b/src/orderings.jl
@@ -1,11 +1,11 @@
 import Base.Order: lt, Ordering
-export LenLex
+export LenLex, BasicWreath
 
 """
     struct LenLex{T} <: Ordering
 
 Basic structure representing Length+Lexicographic (left-to-right) ordering of
-the words over given Alphabet. Lexicographing ordering of an Alphabet is
+the words over given Alphabet. Lexicographic ordering of an Alphabet is
 implicitly specified inside Alphabet struct.
 """
 struct LenLex{T} <: Ordering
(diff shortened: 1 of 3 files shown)


kalmarek:
`WreathOrder` would be better ;)

kalmarek (changes requested):
head does waaaaay too much allocations, but for a stop-gap it's fine;

Maybe you could pick a Wreath-ordered example from Sims and add it to tests as well?

mikolajpabiszczak:
@kalmarek: I saw you already added tests for WreathOrder at some earlier stage. I made them workable by uncommenting etc.

mikolajpabiszczak:
Frankly, I don't know how to shave allocations in head().

kalmarek:
probably by avoiding the `head` completely: Sims describes an algorithm that runs in `O(length(w))` time for producing the sequences. But we don't need the sequences if e.g. the degree of initial words could order them, or the left end;

One possible way is to do this by recursion, i.e. generate the next elements in the sequence only when necessary

mikolajpabiszczak:
Ok, so I written it in the different way. It still allocates, but it allocates half of what it used to allocate.

src/orderings.jl
```diff
@@ -78,16 +57,32 @@ Return whether the first word is less then the other one in a given WreathOrder
...
+        elseif length(head_p) == length(head_q)
+            return @views lt(o, pp[1:first(head_p) - 1], qq[1:first(head_q) - 1])
+        else
+            return false
         end
-        dp < dq ? true : false
+    else
+        return false
     end
-end
```
kalmarek: I'd suggest something like:
```julia
    max_p < max_q && return true
    max_p > max_q && return false
       
    # i.e. max_p == max_q
    head_p_len = count(isequal(max_p), pp)
    head_q_len = count(isequal(max_p), qq)
    
    if head_p_len < head_q_len
        return true
    elseif head_p_len == head_q_len
        first_pp = findfirst(isequal(max_p), pp)
        first_qq = findfirst(isequal(max_p), qq)
        return @views lt(o, pp[1:first_pp - 1], qq[1:first_qq - 1])
    else
        return false
    end
```
note that `findall` allocates an array, which we discard immediately after (we use only its length)

kalmarek (changes requested):

src/orderings.jl
kalmarek: > Ok, so I written it in the different way. It still allocates, but it allocates half of what it used to allocate.
> 
> I also checked the resulting `RewritingSystem` obtained by our implementation for the example 5.5 and two rules are slightly different. I am not sure if this is an error in the book or we have some deeper problem. (Note that this was also the case in the former implementation of `WreathOrder`, the one using `head`).

Thanks for checking this; I don't have time to investigate it now, so either try to understand what goes wrong (probably by means of pen and paper), or open an issue, so that the potential problem is not forgotten.

mikolajpabiszczak: Ech, ofc. this is neat. Updated.

As to the latter: I will open an issue and look into it when I have some time.

kalmarek (approved):

PR #34, opened 2020-10-06
Recursive path ordering

### LINKED ISSUE #32: New examples
this:
https://www.sciencedirect.com/science/article/pii/0022404994900191

defines `RecursivePathOrdering` and applies it to surface groups; Would be great to add it to our tests.

mikolajpabiszczak:
This closes #32. I checked and this ordering produces the right `RewritingSystem` for a system representing the fundamental group of closed orientable surface of genus 3 as reported in [Hermiller, S. M.: Rewriting systems for Coxeter groups](https://doi.org/10.1016/0022-4049(94)90019-1).

As a side note: `LenLex` does not produce a finite system within the limit of `maxrules` for this example, whereas `WreathOrder` does and it is the same system as for `RecursivePathOrder`.

### DIFF
--- a/src/orderings.jl
+++ b/src/orderings.jl
@@ -32,11 +32,11 @@ end
 Base.hash(o::LenLex, h::UInt) = hash(o.A, hash(h, hash(LenLex)))
 
 """
-    lt(o::LenLex, p::T, q::T) where T<:AbstractWord{<:Integer}
+    lt(o::LenLex, p::AbstractWord, q::AbstractWord)
 
 Return whether the first word is less then the other one in a given LenLex ordering.
 """
-function lt(o::LenLex, p::T, q::T) where T<:AbstractWord{<:Integer}
+function lt(o::LenLex, p::AbstractWord, q::AbstractWord)
     if length(p) == length(q)
         for (a, b) in zip(p,q)
             a == b || return isless(a, b)  # comparing only on positive pointer values
@@ -62,11 +62,11 @@ end
 Base.hash(o::WreathOrder, h::UInt) = hash(o.A, hash(h, hash(WreathOrder)))
 
 """
-    lt(o::WreathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
+    lt(o::WreathOrder, p::AbstractWord, q::AbstractWord)
 
 Return whether the first word is less then the other one in a given WreathOrder ordering.
 """
-function lt(o::WreathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
+function lt(o::WreathOrder, p::AbstractWord, q::AbstractWord)
     iprefix = lcp(p, q) + 1
 
     @views pp = p[iprefix:end]
@@ -114,12 +114,12 @@ end
 Base.hash(o::RecursivePathOrder, h::UInt) = hash(o.A, hash(h, hash(RecursivePathOrder)))
 
 """
-    lt(o::RecursivePathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
+    lt(o::RecursivePathOrder, p::AbstractWord, q::AbstractWord)
 
 Return whether the first word is less then the other one in a given
 RecursivePathOrder ordering.
 """
-function lt(o::RecursivePathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
+function lt(o::RecursivePathOrder, p::AbstractWord, q::AbstractWord)
     isone(p) && return !isone(q)
     isone(q) && return false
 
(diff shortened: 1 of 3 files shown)


src/orderings.jl
```diff
@@ -82,4 +93,41 @@ function lt(o::WreathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
...
+    length(p) == 1 && length(q) == 1 && return (first(p) < first(q))
+
+    # i.e. we are not comparing single letter words
+    if first(p) == first(q)
+        @views lt(o, p[2:end], q[2:end]) && return true
+    elseif first(p) < first(q)
+        @views lt(o, p[2:end], q[1:end]) && return true
+    end
+    @views p == q[2:end] && return true
+    @views return lt(o, p[1:end], q[2:end])
```
kalmarek, line 132: this part is a little bit unclear to me; the direct translation of rpo from the paper (as I understand it) would be 
```julia
function lt(o::RecursivePathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
    [...] # deal with exceptional cases here
    # p <_rpo q iff one of the following holds 
    @views begin
        lt(o, p, q[2:end]) && return true
        first(p) < first(q) && lt(o, p[2:end], q) && return true
        first(p) == first(q) && lt(o, p[2:end], q[2:end]) && return true
    end
    return false
end
```
is your version equivalent to this?

src/orderings.jl
```diff
@@ -82,4 +93,41 @@ function lt(o::WreathOrder, p::T, q::T) where T<:AbstractWord{<:Integer}
...
\ No newline at end of file
+end
+
+
+"""
+    struct RecursivePathOrder{T} <: WordOrdering
+
+Structure representing Recursive Path Ordering (determined by the Lexicographic
+ordering of the Alphabet) of the words over given Alphabet. This Lexicographinc
+ordering of an Alphabet is implicitly specified inside Alphabet struct.
```
kalmarek, line 104:
```suggestion
ordering of an Alphabet is implicitly specified inside Alphabet struct.
For the definition see
> Susan M.Hermiller, Rewriting systems for Coxeter groups
> _Journal of Pure and Applied Algebra_
> Volume 92, Issue 2, 7 March 1994, Pages 137-148.
```

kalmarek (changes requested):

src/orderings.jl
mikolajpabiszczak: I believe you are not asking about my `if ... elseif ...` (this works similar to your code, and I agree that your code looks nicer) and rather about why the other condition is outside it and why it has equality first. As to equality: note that the condition (1) in the paper is given as `<=` (or to be precise `>=`).  So in your code we would need to add the `==` condition (and it is indeed needed, as otherwise the ordering will not be compatible with concatenation - I will add a test visualising this as the current one is not enough). Except for that your code should give the same result (with one technical change: as far as I know `@views` is used only when you slice an array - i.e. works for `a[1:end]` but not `a` itself. So we either need to use `a[1:end]` or change `lt(o::RecursivePathOrder, p::T, q::T) where ...` to `lt(o::RecursivePathOrder, p::T, q::S) where ...` - whichever you prefer).

One remark: I did try to understand this ordering and I believe I do not understand it. I do not know how to order the set of words just by looking at it. I know how to check that it is not implemented properly: namely check if it is compatible with concatenation via making sure that subwords of a word are smaller than the word itself.

kalmarek: Oh, that's a good catch with `≥_rpo` in the paper! So the correct version should read:
```julia
    @views begin
        (p == q[2:end] || lt(o, p, q[2:end])) && return true
        first(p) < first(q) && lt(o, p[2:end], q) && return true
        first(p) == first(q) && lt(o, p[2:end], q[2:end]) && return true
    end
```
??

as for the signature of `lt` I'd prefer
```
lt(o::RecursivePathOrder, p::AbstractWord, q::AbstractWord)
```
except for enforcing equality of types, `T` does nothing in the body

mikolajpabiszczak: This is correct or we can just have explicit four conditions like I updated in now:
```julia
    @views begin
        p[1:end] == q[2:end] && return true
        lt(o, p[1:end], q[2:end]) && return true
        first(p) < first(q) && lt(o, p[2:end], q[1:end]) && return true
        first(p) == first(q) && lt(o, p[2:end], q[2:end]) && return true
    end
```
and I think four conditions are even clearer for the reading as they somehow show better (to the reader) that the subwords are `<_rpo` the the whole word.

I actually had the type enforcing because `lt` works only when `AbstractWord{T}` has `T <: Integer` but I looked again and it seems that there is no way for `T` to be something else. So I will update it the way you suggested.

kalmarek (approved):
lgtm!

PR #36, opened 2020-10-18
Kbs automata integration

mikolajpabiszczak:
So this is a very crude integration of Index Automata with Knuth-Bendix procedure (2nd version) - the automaton is updated (i.e. `makeindexautomaton!` is run) every time the `RewritingSystem` is updated. Other changes:

- `Automata` are slightly redesigned (new field `stateslengths` which was moved from the `makeindexautomaton`);
- updated `makeindexautomaton`;
- I think that at some point an error got into `rewrite_from_left` which uses `Automaton` - which I corrected.

I will also add more tests: publishing it so that you can have a look.

And of course a quick benchmark:
```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test


    @btime KnuthBendix.knuthbendix2($R)
    @btime KnuthBendix.knuthbendix2automaton($R)
end;
```

which gives:
```julia
  30.099 ms (22105 allocations: 1.76 MiB)
  11.942 ms (145458 allocations: 11.56 MiB)
```

So even though we update the automaton every time the `RewritingSystem` changes we are faster than normal `kbs2`.

### DIFF
--- a/src/automata.jl
+++ b/src/automata.jl
@@ -145,8 +145,14 @@ initialstate(a::Automaton) = first(states(a))
 noedge(a::Automaton) = a.uniquenoedge
 stateslengths(a::Automaton) = a.stateslengths
 
-Base.isempty(a::Automaton) = isempty(states(a))
+function Base.isempty(a::Automaton)
+    σ = initialstate(a)
+    length(states(a)) == 1 && isone(name(σ)) && return true
+    return false
+end
+
 Base.push!(a::Automaton{N, W}, name::W) where {N, W} = (push!(states(a), State(name, noedge(a))); push!(stateslengths(a), length(name)); a)
+Base.empty!(a::Automaton{N, W}) where{N, W} = (empty!(states(a)); empty!(stateslengths(a)); push!(a, W()); a)
 
 """
     replace(k::NTuple{N, T}, val::T, idx::Integer) where {N, T}
@@ -257,11 +263,13 @@ end
 # Ad index automata
 
 """
-    makeindexautomaton(rws::RewritingSystem, abt::Alphabet)
-Creates index automaton corresponding to a given rewriting system.
+    makeindexautomaton!(a::Automaton, rws::RewritingSystem, abt::Alphabet)
+Takes a given empty automaton (i.e. containing only initial state corresponding
+to empty word) and transforms it into index automaton corresponding to a given
+rewriting system.
 """
-function makeindexautomaton(rws::RewritingSystem, abt::Alphabet=alphabet(ordering(rws)))
-    a = Automaton(abt)
+function makeindexautomaton!(a::Automaton, rws::RewritingSystem, abt::Alphabet=alphabet(ordering(rws)))
+    # a = Automaton(abt)
     # Determining simple paths
     for (idx, (lhs, rhs)) in enumerate(rules(rws))
         isactive(rws, idx) || continue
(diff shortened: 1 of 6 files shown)


src/automata.jl
```diff
@@ -294,7 +297,7 @@ function makeindexautomaton(rws::RewritingSystem, abt::Alphabet=alphabet(orderin
             end
         end
         i += 1
-        indcs = findall(isequal(i), Σᵢ)
+        indcs = findall(isequal(i), stateslengths(a))
```
kalmarek, line 308:
```julia
resize!(idcs, 0)
for (k,len) in enumerate(stateslengths(a))
    len == i && push!(idcs, k)
end
```

kalmarek: also above: I think that `τ = last(walk(a, @view(name(σ)[2:end])))` should cut allocations as well

src/automata.jl
```diff
@@ -307,14 +335,13 @@ to `v`. For standard rewriting `v` should be empty.
...
+    resize!(past_states, length(w) + 1)
     state = initialstate(a)
     past_states[1] = state
-    initial_length = length(v)
 
     while !isone(w)
         x = popfirst!(w)
-        k = length(v) - initial_length + 1
+        k = length(v) + 1
         state = outedges(past_states[k])[x]
```
kalmarek, line 345: why this change? If `v` is non-empty (we're just appending the rewrite procedure to existing `v`) this will fail:
`length(w)` might be small compared to `length(v)` and we get out of bounds

kalmarek (changes requested):

src/automata.jl
mikolajpabiszczak: Ahh, yes that's true. I forgot for a moment about the way interface to rewriting is designed. `k = length(v) - initial_length + 1` will be ok. Sorry for that.

src/automata.jl
mikolajpabiszczak: This suggestion is good, but it has to be applied in line 308 (in line 300 we first declare `indcs`).

mikolajpabiszczak:
And the same benchmark gives now:

```julia
  30.129 ms (22105 allocations: 1.76 MiB)
  11.001 ms (116624 allocations: 9.03 MiB)
```

so there is some gain.

mikolajpabiszczak:
Wrt. our conversation today - another benchmark:
```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test


let R = RWS_Example_237_abaB(6), seed=1
    @btime KnuthBendix.knuthbendix2($R, maxrules=300)
    @btime KnuthBendix.knuthbendix2automaton($R, maxrules=300)
end;
```
and results:
```julia
  2.536 s (217613 allocations: 21.17 MiB)
  281.123 ms (1330914 allocations: 111.83 MiB)
```

Which I think is interesting.

kalmarek:
very nice!

mikolajpabiszczak:
Also tried to test `RWS_Example_237_abaB(8)` - I think at the moment we can call it non-reachable: `maxrules=1000` is not enough.

kalmarek:
and how long does it take for index automata (with 1000 rules?)

mikolajpabiszczak:
```julia
julia> R = RWS_Example_237_abaB(8)
julia> @time KnuthBendix.knuthbendix2automaton(R, maxrules=1000);
└ @ KnuthBendix ~/Documents/GitHub/KnuthBendix.jl/src/automata_kbs2.jl:83
250.637719 seconds (70.67 M allocations: 10.361 GiB, 0.66% gc time)
```

kalmarek (approved):

PR #37, opened 2020-11-10
change the order of args in Base.inv (Alphabet goes first)

kalmarek:
example presentation here:
https://gist.github.com/kalmarek/c2084af80d746807df221a5f4b38cad1

```julia
let (rels, A) = gersten_relations(3, commutative = false), order = LenLex(A), maxr = 1000
    let rws = RewritingSystem(rels, order)
        @time KnuthBendix.knuthbendix2!(rws, maxrules = maxr)
    end

    let rws = RewritingSystem(rels, order)
        @time KnuthBendix.knuthbendix2automaton!(rws, maxrules = maxr)
    end
end
```
timing:
```
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs2.jl:81
  0.246382 seconds (41.12 k allocations: 2.799 MiB)
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/automata_kbs2.jl:83
 35.760633 seconds (15.53 M allocations: 2.042 GiB, 1.18% gc time)
```
i.e. rebuilding the automaton is very expensive, especially when there are lots of words inserted.

Corollary: the other example stresses the performance of rewriting. this one stresses the performance of building/changing the automaton.

I think it'll be better to integrate KnuthBendix with Groups, not the other way round.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,14 @@
 name = "KnuthBendix"
 uuid = "c2604015-7b3d-4a30-8a26-9074551ec60a"
-authors = ["Marek Kaluba <kalmar@amu.edu.pl>"]
+authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Mikołaj Pabiszczak <mikolaj.pabiszczak@gmail.com>"]
 version = "0.1.0"
+
+[deps]
+Groups = "5d8bd718-bd84-11e8-3b40-ad14f4a32557"
 
 [compat]
 julia = "1"
+Groups = "^0.5.2"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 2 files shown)


mikolajpabiszczak:
Nice examples (although somehow not surprising :)).

BTW: it occurred to me today that maybe we should start from making `rewrite_from_left` (the one not using automata) to be able to work in parallel. During rewriting we could, e.g., divide the list of rewriting rules into `number_of_workers` sublists and each worker would try to rewrite the word with one of this sublists (rewrite or perform only single rewrite with the first rule applicable). Now the question: what if multiple workers were able to do some rewriting (do we take the result of the "first" worker or perform another branching and work on many various results of rewriting). There will not be much gain for short words and/or short rewriting systems, but for the long rewriting systems the gain may be huge. Another issue: to check how this design could influence the search of critical pairs (I believe that some solutions may make this search slightly less effective).

kalmarek:
superseded by #38

Issue #25, opened 2020-07-22
Shaving off even more allocations


kalmarek:
Currently (even with #23) the number of allocations in each iteration of `knuthbendix2(rws)`
https://github.com/kalmarek/KnuthBendix.jl/blob/665ad036ecabfc7e37c92c82237c1b8527223aaf/src/kbs2.jl#L80
is proportional to `length(rws)` which is killing the performance in the long run;
We should aim at `O(number of new rules added)`.

But this would need several changes which prompted me to think about the design:

* create the buffers `v::BufferWord{T}` and `w::BufferWord{T}` once per `knuthbendix2` call and pass those through calls `forceconfluence!(...)` → `deriverule!(...)` → `rewrite_from_left!(...)`. Moreover we'd need more than these `v` and `w`, since two subsequent calls to `rewrite_from_left!` happen in `deriverule!(...)`, so more `permanent storage` is needed there to keep the results of rewriting from being overwritten (essentially two `v`s and two `w`s need to be passed through). This feels clumsy and error-prone.
* add those `v`s and `w`s as fields of `rws` which is already passed through the chain. This is very much `kbmag` style :P and has the added "benefit" of not being thread-safe. I'd not go this way.
* borrow the design from hpc solvers, i.e. create a separate `KnuthBendixWork` struct (not a very pretty name, I know:) once (per thread) which encapsulates the needed temporaries, runtime parameters, etc. Then call the function chain above on `KnuthBendixWork`. E.g. of such design is in SCS: https://github.com/cvxgrp/scs/blob/master/include/scs.h.

@bprzybylski @mikolajpabiszczak

kalmarek:
closed by #38

PR #38, opened 2020-11-30
Kbs2 updates

mikolajpabiszczak:
Overview of changes:
 - change the order of args in `Base.inv` (Alphabet goes first) (as in the Pull Request #37);
 - added internal fields to `RewritingSystem` which enable the deletion of the rules during Knuth-Bendix procedure;
 - added another version of kbs2 - deleting rules when (more or less) they become inactive;
 - added `simplifyrule` which simplifies both sides of the rule by getting rid of longest common invertible prefix.

I will upload some benchmarks tomorrow.


mikolajpabiszczak:
Few more remarks:
- `_inactiverules` field in `RewritingSystem` is **not** updated when marking the rule as inactive but must be explicitly updated in `deriverule!()` - this is because right now we have a number of different versions of kbs2: the original one and the one using index automaton do not remove inactive rules, so is it not needed to update this field.
- The Knuth-Bendix procedure which deletes inactive rules (more or less as they occur) is `knuthbendix2delinactive()` - I leave creating a unified interface to calling kbs as a future (final) polish. Right now I think it is better to keep all the main versions of this implementations in this no-much-unified fashion.
- I experimented with adding rules deletion to kbs using index automata, but the benefit of it was not obvious at the very moment. I think it is better to first work and obtain more efficient `updateautomaton!()` procedure and then incorporate rules deletion.

mikolajpabiszczak:
Also benchmarks:

```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test

let R = RWS_Example_237_abaB(6), seed=1
    rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)

    @time rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    @time rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    @time rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)
 end;
```

And the results:
- without rules simplification:
```julia
 2.404575 seconds (217.90 k allocations: 21.187 MiB, 1.60% gc time)
 1.144217 seconds (223.19 k allocations: 21.629 MiB)
 0.295235 seconds (1.34 M allocations: 112.794 MiB, 2.87% gc time)
```

- with rules simplification (used in all 3 functions below):
```julia
 2.031173 seconds (218.03 k allocations: 21.198 MiB)
 1.171759 seconds (224.54 k allocations: 21.769 MiB, 0.35% gc time)
 0.323595 seconds (1.45 M allocations: 121.072 MiB, 2.66% gc time)
```

So in this example simplification of rules in most cases slightly increases the time of execution.

mikolajpabiszczak:
Now another benchmark:

```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test

    rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)

    @time rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    @time rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    @time rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)
 end;
```

And the results:
- without simplification of then rules:
```julia
 0.025590 seconds (22.21 k allocations: 1.761 MiB)
 0.018249 seconds (23.16 k allocations: 1.833 MiB)
 0.018795 seconds (117.55 k allocations: 9.120 MiB, 30.25% gc time)
```
- with rules' simplification:
```julia
 0.025631 seconds (22.21 k allocations: 1.762 MiB)
 0.017883 seconds (23.17 k allocations: 1.834 MiB)
 0.018973 seconds (117.55 k allocations: 9.121 MiB, 32.35% gc time)
```

So here it is hard to say, but it seems like there are situations in which rules' simplification may be beneficial.

mikolajpabiszczak:
I also checked the `gersten_presentation` example:
```julia
let (rels, A) = gersten_relations(3, commutative = false), order = LenLex(A), maxr = 1000
    let rws = RewritingSystem(rels, order)
        @time KnuthBendix.knuthbendix2!(rws, maxrules = maxr)
    end

    let rws = RewritingSystem(rels, order)
        @time KnuthBendix.knuthbendix2delinactive!(rws, maxrules = maxr)
    end

    let rws = RewritingSystem(rels, order)
        @time KnuthBendix.knuthbendix2automaton!(rws, maxrules = maxr)
    end
end
```

The results are for `simplifyrule()` incorporated:
```julia
 └ @ KnuthBendix ~/Documents/GitHub/KnuthBendix.jl/src/kbs2.jl:85
   0.840540 seconds (1.70 M allocations: 83.130 MiB, 2.15% gc time)
 └ @ KnuthBendix ~/Documents/GitHub/KnuthBendix.jl/src/kbs2_with_deleting.jl:20
   0.110126 seconds (242.67 k allocations: 12.308 MiB)
 └ @ KnuthBendix ~/Documents/GitHub/KnuthBendix.jl/src/automata_kbs2.jl:84
  16.599499 seconds (14.92 M allocations: 1.724 GiB, 0.71% gc time)
```

Here clearly rules delation is beneficial.

mikolajpabiszczak:
Also a benchmark for `simplifyrule()`:

```julia
using KnuthBendix
using BenchmarkTools
using Random

A = Alphabet(['a', 'e', 'b', 'p'])
KnuthBendix.set_inversion!(A, 'a', 'e')
KnuthBendix.set_inversion!(A, 'b', 'p')


seed=1
n = length(A)
Random.seed!(seed)
w = Word(rand(1:n, 2000))
v = Word(rand(1:n, 2000))
z = Word(rand(1:n, 2000))

w1 = w * v
w2 = w * z

@btime KnuthBendix.simplifyrule($w1, $w2, $A)
```
and the result:
```julia
 8.624 μs (3 allocations: 8.16 KiB)
```

mikolajpabiszczak:
Ok, so I have a slightly better version of `simplifyrule` the same benchmark as above yields now:
```julia
 7.574 μs (3 allocations: 8.16 KiB)
```
so 1μs better than previous version. (I actually just took out the code from `longestcommonprefix` and combined it with checking for inverse.

kalmarek:
please do so; also please provide the comments explaining `_i`, `_j` etc. fields, e.g. the design notes. Maybe it doesn't make much sense to keep `RewritingSystem` an immutable `struct` after all?

Another idea is taken from #25: maybe we should create `KnuthBendixWork` struct keeping those `_i`, `_j` fields as well as temporary words used in rewriting?

I seems that maybe the best way to use multiple threads is to fix the length of RWS, divide iteration over `(i, j)` into chunks, solve critical pairs in each chunk, mark rules for removal (in each chunk) and only then merge new information (=new rules and rules for removal) from chunks serially (or some kind of merge sort?) `KnuthBendixWork` struct would be usefull then, since each thread could work on one of its own without much interference. Synchronization = merge would be then made in serial manner.

mikolajpabiszczak:
I would go for `KnuthBendixWork` struct. Will do it later this week.

mikolajpabiszczak:
This is not yet finished - I have an idea how to cut even more allocations, but I have to do it carefully.

Benchmarks:

```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test

let R = RWS_Example_237_abaB(6), seed=1
    rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)

    @time rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    @time rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    @time rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)
 end;
```

giving:

```julia
  2.609224 seconds (217.69 k allocations: 21.175 MiB)
  1.169173 seconds (94.19 k allocations: 13.767 MiB)
  0.356291 seconds (1.45 M allocations: 120.720 MiB, 3.49% gc time)
```

(note that most of the changes affect only the middle line in benchmark) - nevertheless it cut a lot of allocations (94k vs. 224k that used to be there).

mikolajpabiszczak:
The latest commit uses and simultaneously closes #25. A benchmark:

```julia
using KnuthBendix
using BenchmarkTools
using Random
using Test

let R = RWS_Example_237_abaB(6), seed=1
    rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)

    @time rws = KnuthBendix.knuthbendix2(R, maxrules=500)
    @time rs = KnuthBendix.knuthbendix2delinactive(R, maxrules=500)
    @time rw = KnuthBendix.knuthbendix2automaton(R, maxrules=500)
 end;
```
yields:

```julia
 2.042057 seconds (217.69 k allocations: 21.175 MiB)
 1.156935 seconds (63.35 k allocations: 11.154 MiB)
 0.382759 seconds (1.27 M allocations: 109.004 MiB, 18.60% gc time)
```

so a slash of 1/3 in comparison with the last benchmark (again: middle time).

mikolajpabiszczak:
I wanted to create a universal interface, but the following did not work:

```julia
function knuthbendix!(rws::RewritingSystem, o::Ordering = ordering(rws);
    maxrules::Integer = 100, implementation = "deletion")

    if implementation == "naive_kbs1"
        knuthbendix1!(rws, o, maxrules)
    elseif implementation == "naive_kbs2"
        knuthbendix2!(rws, o, maxrules)
    elseif implementation == "deletion"
        knuthbendix2deleteinactive!(rws, o, maxrules)
    elseif implementation == "automata"
        knuthbendix2automaton!(rws, o, maxrules)
    else
        @warn("Implementation \"$implementation\" is not defined!")
    end
end

function knuthbendix(rws::RewritingSystem, o::Ordering = ordering(rws);
    maxrules::Integer = 100, implementation = "deletion")

    if implementation == "naive_kbs1"
        knuthbendix1(rws, o, maxrules)
    elseif implementation == "naive_kbs2"
        knuthbendix2(rws, o, maxrules)
    elseif implementation == "deletion"
        knuthbendix2deleteinactive(rws, o, maxrules)
    elseif implementation == "automata"
        knuthbendix2automaton(rws, o, maxrules)
    else
        @warn("Implementation \"$implementation\" is not defined!")
    end
end
```

And frankly I do not know how to make it work.

In general I split everything into files: `derive_rule.jl`, `force_confluence.jl` and `kbs.jl` and proper function are in proper files. I find it slightly cleaner. I also made the methods explicit via dispatch: there is a different method for each of the 4 implementations (kbs1, kbs2 - naive, kbs2 - with deletion, kbs2 - with automata). I consider kbs2 with deletion as default one.

kalmarek:
What exactly failed to work with this implementation? I'd actually suggest something like
```julia
function knuthbendix!(rws::RewritingSystem, o::Ordering = ordering(rws);
    maxrules::Integer = 100, implementation = :deletion)

    impl_list = (:naive_kbs1, :naive_kbs2, :deletion, :automata)
    implementation in impl_list ||
        throw(ArgumentError("Implementation \"$implementation\" of Knuth-Bendix completion is not defined.\n Possible choices are: $(join(impl_list, ", ", " and "))."))

    if implementation == "naive_kbs1"
        return knuthbendix1!(rws, o, maxrules)
    elseif implementation == "naive_kbs2"
        return knuthbendix2!(rws, o, maxrules)
    elseif implementation == "deletion"
        return knuthbendix2deleteinactive!(rws, o, maxrules)
    elseif implementation == "automata"
        return knuthbendix2automaton!(rws, o, maxrules)
    end
end

function knuthbendix(rws::RewritingSystem, o::Ordering = ordering(rws);
    maxrules::Integer = 100, implementation = :deletion)
    return knuthbendix!(deepcopy(rws), o, maxrules=maxrules, implementation=implementation)
end
```
so that we actually don't have to maintain the same list in both functions.
Other than this, this looks ready for merge

kalmarek (dismissed):

mikolajpabiszczak:
Sorry for the delay in commenting: I believe it was something with dispatch. It couldn't find proper methods for calling a function.

I like the interface - this is something I was looking for and couldn't find how to solve that. Nice to see how it should be done. 😅 I allowed myself to introduce the interface to the tests. I feel like this can be merged.

kalmarek (approved):

kalmarek:
Well done!

PR #39, opened 2021-01-20
Doc added and project updated

mikolajpabiszczak:


kalmarek:
ok, I've added some input of mine; Let's focus on finishing #38 for now :)

mikolajpabiszczak:
Btw, I tried to obey the following styling:
- 1 sentence per line;
- lines between 80 and 120 characters;
- breaking of long lines as close to semantic "breaks" as possible.

doc/DesignPrinciples.md
```diff
@@ -0,0 +1,236 @@
...
+
+A field `terminal` is a boolean which indicates whether the state represents a word,
+which is a left-hand side of some rewriting rule in `RewritingSystem`.
+In case this is the case (field `terminal` is `true`) field `rrule` should be equal to the `Word` representing
+the right-hand side of that rewriting rule.
+
+Field `ined` (incoming edges) is a vector of instances of `State` struct - those instances of `State`
+from which there is an edge leading to a state in mention
+(so for a `State` with name "abc" there must be - in particular - a `State` with name "ab" in the `ined` list).
+What is more: always on the first place on that list should be a state representing (having field `name`)
```
kalmarek, line 135: why is this first place so important?

doc/DesignPrinciples.md
```diff
@@ -0,0 +1,236 @@
...
+* `i` field is the iterator over the outer loop and
+* `j` is the iterator over the inner loop
+* `lhsPair` and `rhsPair` are inner `BufferPair`s used for rewriting.
+* `_inactiverules` is just a list of inactive rules in the `RewritingSystem` subjected to Knuth-Bendix procedure.
+
+The aim of this structure is to:
+* enable deletion of inactive rules (which requires updating working indexes `i` and `j` of Knuth-Bendix procedure).
+This deleation is performed by the `removeinactive!(rws::RewritingSystem, work::kbWork)` function which is called
+periodically during certain implementations of Knuth-Bendix procedure.
+* reduce the number of allocations caused by rules rewriting (thus there are two `BufferPair`s).
```
kalmarek, line 222:
```suggestion
* reduce the number of allocations caused by rules rewriting (by maintaining the needed temporary words in `BufferPair`s).
```

doc/DesignPrinciples.md
```diff
@@ -0,0 +1,236 @@
...
+`kbWork{T}` is a helper structure used to iterate over rewriting system in Knuth-Bendix procedure.
+It has the following fields:
+* `i` field is the iterator over the outer loop and
+* `j` is the iterator over the inner loop
+* `lhsPair` and `rhsPair` are inner `BufferPair`s used for rewriting.
+* `_inactiverules` is just a list of inactive rules in the `RewritingSystem` subjected to Knuth-Bendix procedure.
+
+The aim of this structure is to:
+* enable deletion of inactive rules (which requires updating working indexes `i` and `j` of Knuth-Bendix procedure).
+This deleation is performed by the `removeinactive!(rws::RewritingSystem, work::kbWork)` function which is called
```
kalmarek, line 220:
```suggestion
This deletion is performed by the `removeinactive!(rws::RewritingSystem, work::kbWork)` function which is called
```

kalmarek (changes requested):

doc/DesignPrinciples.md
mikolajpabiszczak: It is not a requirement per se - this is just how things are with `makeindexautomaton!` (if you think about how this algorithm works, you will see that this condition is always assured). I think it is nevertheless a condition worth mentioning, as it may be useful if another take at automata will be tried (e.g. automata storing only simple paths).

kalmarek (approved):

PR #42, opened 2021-01-28
Enh/deprecate word()

kalmarek:
initially I was only to replace `W()` by `one(W)`, but then I got carried away...

btw. after spending some time poking in automata I really like the design; I allowed myself to change `noedge` etc. to `failstate` to be closer to nomenclature in the literature.


mikolajpabiszczak (approved):
Very nice! Merging!

Issue #46, opened 2021-04-26
automaton corruption during `knuthbendix!`?


kalmarek:
on `master`:
```julia
function RWS_Example_237_abaB(n)
    Al = Alphabet(['a', 'b', 'B'])
    KnuthBendix.set_inversion!(Al, 'b', 'B')

    a, b, B = Word.([i] for i in 1:3)
    ε = one(a)

    eqns = [b*B=>ε, B*b=>ε, a^2=>ε, b^3=>ε, (a*b)^7=>ε, (a*b*a*B)^n=>ε]

    R = RewritingSystem(eqns, LenLex(Al), bare=true)
    return R
end
```
this example collapses when using automata (but it shouldn't):
```julia
julia> R = RWS_Example_237_abaB(4)
Rewriting System with 6 rules ordered by LenLex{Char}(Alphabet{Char}['a', 'b', 'B']):
   1 ✓ b*B   →  (empty word)
   2 ✓ B*b   →  (empty word)
   3 ✓ a*a   →  (empty word)
   4 ✓ b*b*b     →  (empty word)
   5 ✓ a*b*a*b*a*b*a*b*a*b*a*b*a*b   →  (empty word)
   6 ✓ a*b*a*B*a*b*a*B*a*b*a*B*a*b*a*B   →  (empty word)

julia> aut = KnuthBendix.knuthbendix(R, maxrules=35, implementation=:automata)
Rewriting System with 3 rules ordered by LenLex{Char}(Alphabet{Char}['a', 'b', 'B']):
   1 ✓ a     →  (empty word)
   2 ✓ b     →  (empty word)
   3 ✓ B     →  (empty word)

julia> kbs2 = KnuthBendix.knuthbendix(R, maxrules=35, implementation=:naive_kbs2)
┌ Warning: Maximum number of rules (35) in the RewritingSystem reached.
│                 You may retry with `maxrules` kwarg set to higher value.
└ @ KnuthBendix ~/.julia/dev/KnuthBendix/src/kbs.jl:67
Rewriting System with 36 rules ordered by LenLex{Char}(Alphabet{Char}['a', 'b', 'B']):
   1 ✓ a*a   →  (empty word)
   2 ✓ B*b   →  (empty word)
   3 ✓ b*B   →  (empty word)
   4 ✓ b*b   →  B
   5 ✓ B*B   →  b
   6 ✓ b*a*b*a*b*a*b     →  a*B*a*B*a*B*a
[...]

PR #51, opened 2021-08-09
Taking Rules seriously

kalmarek:
There are lots of changes here. Most notable:

1. introduce native `Rule` struct that localizes the `isactive` information → no need to keep track of `i`, `j` and what not
2. iteration over rules of `rws` is storage agnostic now. returned rules are `active` at the time of iteration
3. `knuthbendix2` uses (a much simplified) `kbWork` by default now
4. new `Rule`s are created (and oriented) only when a pair turns out to be critical. This saved allocations MASSIVELY.
5. `rewrite_from_left!(v,u,rws)` semantic was changed: we no longer append to `v`, we write over its storage.
6. Introduce `ProgressMeter` gimmick ;)

Minor changes:
* `isprefix` and `issuffix` don't take additional argument `k` (aka `length`). passing `@view w[1:k]` is more readable and as fast as the old version
* added basic `balancerule!` (used only at rws creation at the moment)
* `RWS_Example_6_4` and `6_5` were fixed and existing `@test_broken` removed. However #46 is still present.
* `RewritingSystem` constructor now makes sure that rules are properly oriented upon creation.

fixes:
* `getirreduciblesubsystem` was fixed by introducing `subwords` iterator
* `simplifyrule!` failed when common prefix and common suffix overlap
* bunch of other small stuff

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "KnuthBendix"
 uuid = "c2604015-7b3d-4a30-8a26-9074551ec60a"
 authors = ["Marek Kaluba <kalmar@amu.edu.pl>", "Mikołaj Pabiszczak <mikolaj.pabiszczak@gmail.com>"]
-version = "0.2.1"
+version = "0.3.0"
 
 [deps]
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.2.1"
 
 [deps]
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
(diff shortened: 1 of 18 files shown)


kalmarek:
I will think about unifying different implementations since for now the whole main loop looks like this:
```julia
function knuthbendix(rws::RewritingSystem; kwargs...)
    # initialize rws here, e.g. creating kbWork, stack, etc
    objs = ...
    
    for ri in rules(rws)
        # check early breaking conditions here
        for rj in rules(rws)
            isactive(ri) || break
            forceconfluence!(rws, ri, rj, objs... )
            isactive(rj) || break
            ri == rj && break
            forceconfluence!(rws, rj, ri, objs... )
        end
        remove_inactive!(rws, objs... )
    end
    return rws
end
```

PR #67, opened 2023-09-21
Benchmarking

kalmarek:
@RearrangedLetters I decided to take those changes that you made and turn them into benchmarking PR on which parallelization builds itself; 

I will clean this up and then merge on in the next week or so.


kalmarek:
The single entry point to benchmark is `benchmarking/run_benchmarks.jl` which produces time-stamped `csv` files from dataframes and the comparison plot in `benchmarking/results`.

I'm happy with how things look now. Thanks @RearrangedLetters for laying the foundations of this PR!

